PR #10, opened 2018-10-21
Preliminary oneshot support!

nathunsmitty:
Uses `keybase oneshot` instead of relying on current login credentials.

Note: I was having some trouble with asynchronous events and logging out on completion, so I left that out for now. Current behavior is that you'll continue to run in one-shot mode after a script executes. This is something we should change, although I think we could probably get this after we promisify things.

Other miscellanea:
- I updated `babel-eslint` because I was getting an error and also add a couple of rules.
- I think we may want to ignore index.js for git, prettier, eslint, etc. since it's autogenerated
- Lots of things got prettified!


lib/bot.js
```diff
@@ -34,23 +45,25 @@ class Bot {
...
-      if (currentDPair) {
-        this._dPair = currentDPair
-        this._log(`intialized ${currentDPair.username} (device=${currentDPair.devicename})`)
-      }
-      this._initialized = true
-      cb(err)
+  init(options: InitOptions, cb: CbError): void {
+    this._verbose = options.verbose
+
+    keybaseExec({args: ['oneshot', '--username', options.username, '--paperkey', options.paperkey]}, () => {
```
thebearjew: This doesn't handle `err` from `keybaseExec`. I was able to send a chat message without providing any options (username or paperkey).

PR #13, opened 2018-11-07
Update Flow Typing

thebearjew:
**README**
1. Update the examples to correctly use the Promise API
2. Update example options to show missing options and optional parameters
3. Add `chatOptionsList` to README instead of null

**lib/bot.js**
2. Add Flow typing for chat command options

### DIFF
--- a/lib/types.js
+++ b/lib/types.js
@@ -5,169 +5,169 @@
  * Source: https://github.com/keybase/client/blob/master/go/client/chat_api_handler.go
  */
 
-export type ChatChannel = {
+export type ChatChannel = {|
   name: string,
   public: boolean,
   members_type: string,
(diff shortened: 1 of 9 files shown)


lib/types.js
```diff
@@ -1,35 +1,173 @@
...
+  sent_before?: string,
+  before_contex?: number,
+  after_context?: number,
+}
+
+/*
+ * List
+ * {"method": "list", "params": {"options": { ...ChatOptionsList }}}
+ */
+export type ChatOptionsList = {
```
nathunsmitty: It might make more sense to make most of these types exact object types

thebearjew: Agreed!

Issue #18, opened 2018-11-09
Update to a general keybase bot, not just a chat bot


nathunsmitty:
- [ ] Rename module to `keybase-bot`
- [ ] Bump module version to 2.0
- [ ] Rename method calls to allow future expandability, e.g., `bot.chatSend` -> `bot.chat.send`. I think the following modules would be a good starting point:
```
chat
files
teams
git
wallets
encrypt
decrypt
```
For this, everything but `chat` can be stubbed for now and we can worry about implementation later.

thebearjew:
@nathunsmitty I think now is a good time to begin restructuring the module. Might as well break chat into its own package at the same time as we change the method signatures (previous comment) and do required argument checking (#16).

PR #28, opened 2018-11-10
Add a `deinit` bot method

nathunsmitty:
With the introduction of one-shot mode, there wasn't a way to log the bot out after `bot.init()` was run. I think the best solution to this problem is to add a `bot.deinit()` method that should be run at the end of every bot program.

Other changes:
- Added/changed documentation (both in readme and jsdoc) to reflect the `bot.deinit()` addition
- Added a VSCode debug task for math bot 🤖 
- Fixed some issues with math bot, but I think more rigorous testing is needed. I stop it from crashing, but I'm getting an issue where it's looking at all messages in a chat instead of just new ones. It seemed a little outside the scope of this PR, but would probably be something good to look at when we restructure the chat stuff as per #18 
- Changed the examples to use async/await, which I think fit the `deinit()` paradigm better. General bot usage now looks like:
```javascript
try {
  await bot.init({...})
  ....
} catch (error) {
  console.error(error)
} finally {
  bot.deinit()
}
```
- Assorted prettier/eslint autofixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,14 +63,17 @@ bot
 
 ### Commands
 
-
 Anywhere a promise is returned from the bot API, you should handle error with `.cathc(err => { ... })`
 
-#### `bot.init(options, cb)`
+#### `bot.init(options)`
 
 As shown above, this must be run to initialize a bot before using it. It checks to make sure you're properly logged into Keybase and gets basic info about your session. Afterwards, feel free to check bot.myInfo() to see or check who you're logged in as.
 
 `options` is a dictionary expecting `username` and `paperkey` (which are pretty self-explanatory), as well as `verbose`, which says whether the bot should log much of what it's doing.
+
+#### `bot.deinit()`
+
+This should be run to after your bot finishes running. It logs you out of your Keybase session.
 
 #### `bot.myInfo()`
 
@@ -90,7 +93,6 @@ lists your chats, with info on which ones have unread messages.
   fail_offline?: boolean,
 }
 ```
-
 
 #### `bot.chatSend(chatOptionsSend) : Promise`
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -21,57 +21,43 @@ npm install keybase-chat-bot
...
+}
 ```
 
 ### Commands
 
-
 Anywhere a promise is returned from the bot API, you should handle error with `.cathc(err => { ... })`
 
-#### `bot.init(options, cb)`
+#### `bot.init(options)`
```
thebearjew: Rebase on master for the `options` typing in the README

lib/full-watcher.js
```diff
@@ -83,7 +83,7 @@ class FullWatcher {
   // --------------------------------------------------------------------------
 
   _watchLoop(): void {
-    this._checkForNewMessagesInAllConversations.then(() => {
+    this._checkForNewMessagesInAllConversations().then(() => {
```
thebearjew: Thanks for making this fix!

lib/bot.js
```diff
@@ -42,8 +45,13 @@ class Bot {
...
 
-  // --------------------------------------------------------------------------
-
+  /**
+   * Initializes a bot by logging in with the given credentials. This must be run before your bot can be used.
+   *
+   * @param {InitOptions} options – An object expecting `username` and `paperkey` (which are pretty self-explanatory), as well as `verbose`, which says whether the bot should log much of what it's doing.
+   * @returns {Promise<?string>}
+   * @memberof Bot
+   */
```
thebearjew: Quick thought on documentation @nathunsmitty.

We're already flow typing. Perhaps keeping the JSDOC syntax, but switching to a build system like [documentation.js](https://github.com/documentationjs/documentation) might be helpful as it supports reading flow types and can get `@param` and `@return` correctly.

I'll make a issue to research how well they jive together.

PR #33, opened 2018-11-16
Use chat listen api instead of polling

### LINKED ISSUE #6: Use listen mode
If a bot author wants to watch for new messages, our bot shouldn't poll. Instead it should make a call to keybase chat api in listening mode. Which just streams lines of JSON when there's new chat activity. This will run faster and not use continuous "gas" when there's no activity.

nathunsmitty:
Yay cleaner code!

closes #6


lib/bot.js
```diff
@@ -55,15 +43,17 @@ class Bot {
...
-    return keybaseExec(['oneshot', '--username', options.username], options.paperkey).then(() => {
-      return getKeybaseUsernameAndDevicename().then(currentDPair => {
-        if (currentDPair) {
-          this._dPair = currentDPair
-          this._log(`intialized ${currentDPair.username} (device=${currentDPair.devicename})`)
-        }
-        this._initialized = true
-      })
-    })
+    return keybaseExec(['oneshot', '--username', options.username], {stdinBuffer: options.paperkey}).then(
```
thebearjew: 👍

lib/bot.js
```diff
@@ -106,44 +88,47 @@ class Bot {
...
-    if (this._channelWatchers.has(key)) {
-      throw new Error(`already watching the channel ${JSON.stringify(channel)}`)
-    } else {
-      this._channelWatchers.set(key, new ChannelWatcher({channel, onMessages, bot: this}))
-    }
+  /**
+   * Listens for new chat messages on a specified channel. The `onMessage` function is called for every message your bot receives.
+   * @param onMessage
+   * @memberof Bot
+   */
```
thebearjew: Just a quick heads up on the JSDOC stuff. I'm currently finishing the documentation build on `thebearjew/documentation-js`.

The approach is to move all the explanations & examples that _were_ in the README into JSDOC comments above each method. More than likely, these comments will be replaced in favor of the README comments.

README.md
```diff
@@ -174,53 +174,44 @@ Known bug: the GUI has a cache, and deleting from the CLI may not become apparen
 
 #### `bot.watchAllChannelsForNewMessages(options)`
```
thebearjew: heads up on the README here. I'm working on `thebearjew/documentation-js` which should build the method signatures, types, descriptions, and examples from the JSDOC comments in the source code.

lib/bot.js
```diff
@@ -106,44 +88,55 @@ class Bot {
...
+  watchChannelForNewMessages(channel: ChatChannel, onMessage: (message: ChatReadMessage) => void) {
+    keybaseExec(['chat', 'api-listen'], {
+      onStdOut: chunk => {
+        try {
+          const streamString = chunk.toString('utf8')
+          const messageObject = JSON.parse(streamString)
+          if (channel.name === messageObject.msg.channel.name) {
```
nathunsmitty: Not super happy with how we filter messages by channel, since we don't match impteam names that aren't in the canonical order. I think this is fine for now, but requires a larger discussion.

nathunsmitty:
this closes #6

lib/bot.js
```diff
@@ -106,44 +88,47 @@ class Bot {
...
+  /**
+   * Listens for new chat messages on a specified channel. The `onMessage` function is called for every message your bot receives.
+   * @param onMessage
+   * @memberof Bot
+   */
+  watchChannelForNewMessages(channel: ChatChannel, onMessage: (message: ChatReadMessage) => void) {
+    keybaseExec(['chat', 'api-listen'], {
+      onStdOut: chunk => {
+        const streamString = chunk.toString('utf8')
+        const messageObject = JSON.parse(streamString)
```
thebearjew: Is it safe to assume that the `chunk` size will be the full JSON text? What if the JSON is larger than the chunk size?

The number of bytes read is defaulted to 16Kb for Readable streams: https://nodejs.org/api/stream.html#stream_new_stream_readable_options

But that depends on how `spawn` initializes the `Stream.Readable` for `STDOUT`.

This might not be an issue in testing, but I'm curious if it's possible to get a `JSON.parse` error, which reminds me that it should be wrapped in `try { ... } catch (e) { ... }`

nathunsmitty:
@thebearjew chunk size problem is fixed with `readline` – should be ready for another look at!

thebearjew:
@nathunsmitty Merged documentation into this branch

demos/math_bot.js
```diff
@@ -11,12 +11,11 @@ const mathjs = require('mathjs')
...
 //
 
-const msgReply = function(s) {
+const msgReply = s => {
   let a1, a2, ans, b1, b2, eqn
   try {
     ans = '= ' + mathjs['eval'](s).toString()
-  } catch (error) {
-    console.error(error)
+  } catch {
```
thebearjew: This is invalid JavaScript syntax, throwing an error.

Needs to be

```js
try { ... } catch (error) { ... }
```

I'm running `lts/carbon → 8.12.0.` and it's still invalid.

PR #34, opened 2018-11-16
Generate API Documentation With Flow Types

thebearjew:
1. Adds [documentation.js](https://github.com/documentationjs/documentation)
2. Use Flow types and JSDOC comments to build API markdown docs
3. Changes all chat options to use `camelCase` for a more consistent JavaScript API.
4. Convert all `camelCase` option keys into `snake_case` before passing to `keybase chat api`
5. Adds `yarn docs` to the package scripts 
6. Rebased on `nathan/api-listen` to include method changes (until that branch is merged)

Closes #30

PR #35, opened 2018-11-18
[WIP] Restructure the Bot(s)

### LINKED ISSUE #18: Update to a general keybase bot, not just a chat bot

nathunsmitty:
🚨WIP: demos and api seem functional, but are not currently well documented or typed and definitely don't robustly handle errors yet. Wanna get around to that after we merge #34 and #33 in. 🚨

This PR restructures the bot to match a more general api, of the format `bot.chat.send`. To do this, I thought the best way to structure the bot was as a collection of more specific bots.

All bots inherit from a `BaseBot` class, which provides general methods for initialization, reinitialization, logging, etc. Specialized bots, such as `ChatBot`, extend this functionality with their own methods and helpers. Our general `Bot` class also inherits from `BaseBot`, and contains each specialized bot as a property on it.

I also tried to use async/await features over promises where it made sense, as I think this results in easier to understand code. I ran into some issues with async class methods, which led to some babel/flow reconfiguring.

I also tried to simplify the number of files we had. All chat logic went into `ChatBot`. `execToJSON` was removed in favor of `keybaseExec`, etc.

closes #18


demos/check_for_unreads.js
```diff
@@ -6,12 +6,16 @@ async function main() {
 
   try {
     await bot.init({username: process.env.KB_USERNAME, paperkey: process.env.KB_PAPERKEY, verbose: false})
-    const res = await bot.chatList({
+    const res = await bot.chat.list({
       showErrors: true,
-      unreadOnly: true,
+      unreadOnly: false,
```
nathunsmitty: I think this should be `false`, since we filter the unreads and compare them to the total

PR #39, opened 2018-11-22
Remove execToJSON; add a more explicit JSON parsing option to keybaseExec

nathunsmitty:
`keybaseExec` was originally created to handle the case of output from commands not being JSON. However, since it shares a lot of code with `execToJSON`, it doesn't make sense to have both. This adds a more explicit JSON parsing option to `keybaseExec` (it kinda implicitly parsed JSON previously), and updates the places that were previously using `execToJSON`.


lib/chat-api.js
```diff
@@ -22,30 +22,14 @@ const runApiCommand = (arg: ApiCommandArg): Promise<any> => {
...
-          reject(new Error(oError.message || oError.toString()))
-        } else {
-          reject(new Error(`Unknown error parsing result - no "result" field`))
-        }
-        resolve(res)
-      })
-      .catch((err?: Error) => {
-        reject(err)
-      })
+  const inputString = JSON.stringify(input)
```
nathunsmitty: Not an expert on error handling with promises and async functions, so I'm not sure if we need a try catch for errors here. Since the previous version was just throwing errors up, I thought no but I'm not sure.

lib/keybase-exec.js
```diff
@@ -5,12 +5,13 @@ import readline from 'readline'
 type ExecOptions = {|
   stdinBuffer?: Buffer | string,
   onStdOut?: (line: string) => void,
+  parseOutputAsJSON?: boolean,
```
nathunsmitty: If we're parsing JSON a lot (which we will be for chat), we could make this a negative. I suppose we could also leave the JSON parsing to be implicit, but that almost feels too magical to me.

lib/keybase-status.js
```diff
@@ -1,36 +1,14 @@
...
-        resolve({username: status.Username, devicename: status.Device.name})
-      } else {
-        reject(new Error('failed to get username + device name'))
-      }
-    })
-  })
+import keybaseExec from './keybase-exec.js'
+import type {UsernameAndDevice} from './types.js'
+
+const getKeybaseUsernameAndDevicename = async (): Promise<UsernameAndDevice> => {
```
nathunsmitty: I don't think it really makes sense to keep `getKeybaseNativeStatusJson`; we weren't using it anywhere except in the same file

lib/keybase-exec.js
thebearjew: I see what you mean here.
Quick thing - perhaps change this flag to just `json?: boolean` to reduce the function signature?

Since the chat API methods are funneled through `runApiCommand` we can always pass `json: true` right? That basically sets it for every call to `keybase chat api`. Also no one using this library will be calling `keybaseExec` directly so this shouldn't be too much of an issue.

I think it's safer to leave the flag as a positive because it would be more confusing to set the default to `true`, have someone forget to turn it off, parse the non JSON output as JSON, get a `JSON.parse()` error` and then be confused.

lib/keybase-status.js
thebearjew: Nuke it

lib/types.js
```diff
@@ -178,4 +178,4 @@ export type ChatReadMessage = any
 export type DeviceUsernamePair = {|username: string, devicename: string|}
 /**
  */
-export type NativeStatus = void | {Device?: {name: string}, Username?: string}
+export type UsernameAndDevice = {|username: string, devicename: string|}
```
thebearjew: `keybase status -j` can fail, return nothing, return non-json (message to STDOUT), etc. 

What we're doing on `keybase-status.js#L7` is handling the falsy fields to convince Flow that the branch of execution to `return` would not happen unless both fields are truthy. So these properties are very much still optional and should stay that way in the Flow typing. One option is to split this into two different types:

1. `MaybeUsernameAndDevice` for the output from `keybase status -j`
2. `UsernameAndDevice` for the resolved type of the Promise

nathunsmitty: If `keybase status -j` doesn't return a valid username and devicename, won't we just throw an error in which case it doesn't matter what our function returns? We only get `UsernameAndDevice` if we have both a username and device from `keybase status`, right?

thebearjew: @nathunsmitty Good point. The previous implementation handled the case of `keybase status -j` succeeding (exit code 0) and the username / device not being present in the output.

As you mentioned, it might be case that the service will always either succeed with valid JSON or error (exit code != 0) and the previous implementation was against a different `keybase status` behavior.

I think you're right about this - I'll take a look at the code and ping someone one CORE to double check. For now this works, thanks for pointing this out!

thebearjew: Note to self: **Write a test for this**

PR #42, opened 2018-11-24
Don't just get unread conversations in `check_for_unreads` demo

nathunsmitty:
Since the demo filters unread conversations, we don't want just the unread messages from the service. It looks like the unread only option was only introduced when we updated the flow types #13.

I only realized this would be an issue when I ran this demo while having no unread messages – I got a type error for trying to run `filter` on `null`. Even with unread messages, though, the results aren't correct as `res.conversations.length` doesn't give the number of conversations, but just the ones with unread messages.

I also added a check in the rare case someone runs this on an account where there are no conversations (and `res` is thus `null`).

### DIFF
--- a/demos/check_for_unreads.js
+++ b/demos/check_for_unreads.js
@@ -8,7 +8,6 @@ async function main() {
     await bot.init({username: process.env.KB_USERNAME, paperkey: process.env.KB_PAPERKEY, verbose: false})
     const res = await bot.chatList({
       showErrors: true,
-      unreadOnly: true,
     })
     const unreadCount = res.conversations.filter(c => c.unread).length
     console.log(`You have ${unreadCount} unread conversations out of ${res.conversations.length} total`)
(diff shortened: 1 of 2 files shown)


demos/check_for_unreads.js
```diff
@@ -8,10 +8,13 @@ async function main() {
     await bot.init({username: process.env.KB_USERNAME, paperkey: process.env.KB_PAPERKEY, verbose: false})
     const res = await bot.chatList({
       showErrors: true,
-      unreadOnly: true,
     })
-    const unreadCount = res.conversations.filter(c => c.unread).length
-    console.log(`You have ${unreadCount} unread conversations out of ${res.conversations.length} total`)
+    if (res) {
```
thebearjew: @nathunsmitty good catch on accounts without conversations.

Here's a thought. Why not maintain our return type that `chat.list` will always return an array and default to an empty array in `chatList` (what will soon be `chat.list`).

So if the return value from `keybase chat list` is `null` or `undefined`, just return empty list `[]`.

This was having an inbox with no unreads can be handled the same as the case where there are no conversations in the inbox at all.

PR #43, opened 2018-11-25
🤓 Functional Restructure

thebearjew:
### Goals

The purpose of restructuring is to transition the module into a general keybase bot as outlined in #18. The approach taken draws from functional programming with the hopes of improving usability, testing, and development.

- Improve future unit + integration testing by refactoring each API into functions rather than classes
- Make bot state transitions predictable and pure with a "redux"-like store
- Writing tests for multiple concurrent bots should be easy.
- Spinning up a bot into any possible state combination should be easy

### Overview

1. Define `chat/` `base/` and `team/` modules
2. Define a state solution, `store/` which uses pure functions to update state
    - `createStore(reducer, initialState)`
    - `store.getState()`
    - `store.dispatch(action())`
    - `withStore()`
    - `withGuards()`
3. Define function `withGuards([guardFn1, guardFn2])` that takes an arbitrary number of guard functions and waits for all to resolve.
4. Define a helper `withState(store, withGuards(guards))(wrappedApiFunc)` which composes and `apiFunc` to be able to use `store` without changing its function signature. If `guards` is provided, then all guards must resolve before calling `apiFunc`.
5. Lots of Flow type updates for better typing
6. Update `rollup.config.js` to use Node/NPM modole resolution - this means that it will use `index.js` as the default export of a directory.
7. Update JSDOC syntax to use `@namespace` to structure module hierarchy.

### Bot API

The Bot API is mostly the same but included the discussed changed in #18.

```js
const Bot = require('keybase-chat-bot')
const bot = new Bot()
// Base
bot.init({username: '', paperkey: '', verbose: false})
bot.myInfo()
bot.deinit()

// Chat
bot.chat.list( ... )
bot.chat.read( ... )
bot.chat.send( ... )
bot.chat.delete( ... )
bot.chat.watchChannelForNewMessages( ... )
bot.chat.watchAllChannelsForNewMessages( ... )
```

However when Bots are initialized they create a functional `store` and pass that to each API function using `withStore`.
I kept object instantiation to maintain the same API for now. The alternative would have been `const newBot = require('keybase-chat-bot')` and that seemed a bit off.

### Tests

With API functions being able to have closure over `store`, we will be able to write tests that do the following:

1. Test a bot in any state by defining the state as an object
2. When running tests we can test any API function without having to call `init()` first because we can use `withStore()` and pass a testing store in.
3. Start with any testing store, run `N` API commands, and check to make sure the store is correct
4. Create two stores, two bots that communicate together,  and run `N` API commands on bot 1 and `N` API commands on bot 2 and check to make sure that the store is correct.


lib/types.js
```diff
@@ -170,12 +199,60 @@ export type AllChatOptions =
...
+
+export type MyInfo = () => ?DeviceUsernamePair
+
+export type List = ChatOptionsList => Promise<void | ChatConversation[]>
+
+export type Read = ChatOptionsRead => Promise<any>
+
+export type Send = ChatOptionsSend => Promise<void>
+
+export type Delete = ChatOptionsDelete => Promise<void>
```
nathunsmitty: How do these types relate to the function type signatures in `chat/index.js`? `deleteMessage` has a return type of `Promise<any>` over there. These look to be the more authoritative ones since they're used in the "with store" functions, but does that mean we have to duplicate function signatures?

package.json
```diff
@@ -64,6 +64,7 @@
     "mathjs": "^5.2.3",
     "prettier": "^1.15.2",
     "rollup": "^0.67.3",
-    "rollup-plugin-babel": "^4.0.3"
+    "rollup-plugin-babel": "^4.0.3",
+    "rollup-plugin-node-resolve": "^3.4.0"
```
nathunsmitty: We can actually remove `rollup-plugin-node-resolve` – it bundles dependencies into the bundle. I think we'd rather just declare our dependencies and dependencies and have npm/yarn handle downloading those. In our case, I guess it's pretty small since we're only doing the lodash snakecase package.

lib/chat/index.js
```diff
@@ -0,0 +1,177 @@
...
+export const send = async (options: ChatOptionsSend): Promise<void> => {
+  const res = await _runApiCommand({method: 'send', options})
+  if (res.hasOwnProperty('error')) {
+    throw new Error(res.error.message)
+  }
+}
+
+/**
+ * Deletes a message in a channel. Messages have messageId's associated with
+ * them, which you can learn in `bot.chatRead`. Known bug: the GUI has a cache,
```
nathunsmitty:
```suggestion
 * them, which you can learn in `bot.chat.read`. Known bug: the GUI has a cache,
```

lib/bot.js
```diff
@@ -1,221 +1,26 @@
 // @flow
-import getKeybaseUsernameAndDevicename from './keybase-status.js'
-import {runApiCommand} from './chat-api.js'
-import keybaseExec from './keybase-exec'
```
nathunsmitty: We can delete the `keybase-exec` file too! 🔥

lib/base/index.js
```diff
@@ -0,0 +1,115 @@
...
+import {keybaseExec} from '../utils'
+import {withStore, actionInit} from '../store'
+import type {Store} from '../store/types'
+import type {InitOptions, DeviceUsernamePair, UsernameAndDevice} from '../types.js'
+
+/**
+ * @ignore
+ * _getCurrentUsernameAndDevicename returns { username, devicename } from keybase status -j
+ */
+const _getCurrentUsernameAndDevicename = async (): Promise<UsernameAndDevice> => {
```
nathunsmitty: Since most (all?) underscore methods in Bot are now pure functions 😍 can we get rid of the underscores? No need for hacky JS object-oriented stuff!

thebearjew: Do underscored functions do something in classes? I always thought it was a convention for denoting private methods desire JavaScript not having private methods.

lib/types.js
```diff
@@ -1,12 +1,32 @@
...
+/**
+ */
+export type NativeStatus = void | {Device?: {name: string}, Username?: string}
+
+export type ApiCommandArg = {method: string, options: Object}
 
 /**
  */
-export type InitOptions = {|
+export type InitOptions = {
```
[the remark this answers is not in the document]
thebearjew: We don’t! This was he result of wrangling a nasty flow error and I didn’t revert.

package.json
thebearjew: @nathunsmitty hmm, I'm pretty sure it just uses the algorithm to resolve and doesn't necessarily include the modules? Just looked over the repo again, but I might be missing something.
https://github.com/rollup/rollup-plugin-node-resolve

lib/base/index.js
nathunsmitty: Oh yeah, the underscores are just a naming convention. But since `_getCurrentUsernameAndDevicename` isn't a private method anymore, not really sure it makes sense.

lib/bot.js
nathunsmitty: Looks like it is now – may have been on the wrong commit or branch.

lib/store/types.js
```diff
@@ -0,0 +1,33 @@
...
+
+export type DeinitAction = {|
+  type: 'DEINIT',
+  payload: {||},
+|}
+
+export type Action = InitAction | DeinitAction
+
+export type State = {|
+  +initialized: boolean,
```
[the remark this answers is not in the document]
nathunsmitty: very cool!

package.json
nathunsmitty: Turns out we're both right – we need `rollup-plugin-node-resolve` to resolve `index.js` files but it also will try to resolve modules in `node_modules`. To avoid this, we have to use the `externals` option in the rollup config, which we're already doing so I think this is good 👍

nathunsmitty:
@thebearjew

package.json
thebearjew: Hooray! Thanks for digging into this!

lib/store/index.js
```diff
@@ -0,0 +1,141 @@
...
+    state = reducer(state, action)
+  }
+  return {
+    getState,
+    dispatch,
+  }
+}
+
+/*
+ * Actions
```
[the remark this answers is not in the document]
thebearjew: Just split these out to `actions.js` and `reducers.js` for now because things will be simple for the foreseeable future.

lib/types.js
```diff
@@ -16,147 +36,156 @@ export type InitOptions = {|
 type TopicType = 'chat' | 'DEV'
 type MembersType = 'team'
 
-export type ChatChannel = {|
+export type ChatChannel = {
```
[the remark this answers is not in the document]
thebearjew: These types have been updated and split into `chat/types.js` `base/types.js` `./types.js`

lib/types.js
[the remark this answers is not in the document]
thebearjew: Updated!

demos/check_for_unreads.js
```diff
@@ -1,17 +1,17 @@
...
-
   try {
     await bot.init({username: process.env.KB_USERNAME, paperkey: process.env.KB_PAPERKEY, verbose: false})
-    const res = await bot.chatList({
-      showErrors: true,
-      unreadOnly: true,
-    })
-    const unreadCount = res.conversations.filter(c => c.unread).length
-    console.log(`You have ${unreadCount} unread conversations out of ${res.conversations.length} total`)
+    const conversations = await bot.chat.list({showErrors: true, unreadOnly: false})
```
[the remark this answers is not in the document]
thebearjew: Reverted back to version on master

thebearjew:
@nathunsmitty Made the suggested changes. Also took the liberty to make a breaking API change by moving to direct function arguments instead of options.

**Before**
```js
bot.chat.send({channel, message, ...etc })
```

**After**
```js
bot chat.send(channel, message, {...etc})
```

lib/bot.js
```diff
@@ -1,221 +1,25 @@
...
- * A Keybase bot.
- *
- * @class Bot
- */
 class Bot {
-  _dPair: ?DeviceUsernamePair
-  _initialized: boolean
-  _verbose: boolean
+  store: Store
+  init: any
```
nathunsmitty: Is it possible to enforce types here? For example, I think `init` should have a type of `InitOptions => Promise<void>`, at least according to the docs.

I get that `withStore` kinda adds a lot of complexity to the typing, though, and I'm sure you've looked into this.

lib/chat/index.js
```diff
@@ -0,0 +1,193 @@
...
+  ChatListOptionals,
+  ChatSendOptionals,
+  ChatReadOptionals,
+  ChatDeleteOptionals,
+  ChatConversation,
+  ChatChannel,
+} from './types'
+
+const runApiCommand = async (arg: ApiCommandArg): Promise<any> => {
+  // await guardInitialized()
```
nathunsmitty:
```suggestion
```
A leftover from a previous idea?

lib/bot.js
thebearjew: @nathunsmitty Definitely on the same page here. I was wrangling types for a while as I was writing `withStore`. It's tough to get Flow to recognize different possible types in a `Union` type when each one is a function definition.

I might go back to the drawing board on this soon. I'll make a separate ticket for expanding typing to these places.

lib/chat/index.js
thebearjew: Fixed

PR #49, opened 2018-12-06
Coyne/index and npm ignore

malgorithms:
hi @nathunsmitty  and/or @thebearjew, curious your thoughts on this.

(1) I introduced a `.npmignore` file, so our module will be as lightweight as possible to install, not pulling along all the random config files and demos. It shrinks the package by like 70% when I do an `npm publish --dry-run`. Note this still includes `lib/` which I assume we want along for the ride since `index.js.map` points at it.

(2) I reintroduced `index.js` in the repo, even though it's autogenerated. I think it's convenient for people looking at the module on github to see the code; otherwise they look at package.json, see the reference to index.js, but can't find it, and need to understand our build process.  As a second bonus, it allows us or someone else to check out a branch and do an `npm link` to try it, without going through build steps.


nathunsmitty:
Instead of an `.npmignore` file, it may fit this project better to use the [files field](https://docs.npmjs.com/files/package.json#files), which acts as a whitelist. I think the only files and folder we would include are `lib/`, `index.js`, and `index.js.map`? It's probably likely that we add additional dot/config/rc files to the project than change the locations of these files. Whether we go the black- or whitelist route, though I support making the module as lightweight as possible.

I'm not sure I'm a fan of readdinng `index.js` and `index.js.map` back into source control though. The reason we originally removed it was because it leads to a **lot** of merge conflicts and noise on commits, as every single source code change will affect the contents of both files. I think most people browsing through our code on GitHub would be more interested in the source code instead of the transpiled version, and we could clear up any confusion by mentioning `index.js` is autogenerated in the README.

To meet that second point, I think we should include a prepublish script that runs our build script. [Prepublish](https://docs.npmjs.com/misc/scripts) runs when a user runs `npm install` or `yarn install`, so that way a user who clones our repo can have `index.js` without any additional steps after install dependencies. (It also ensures that `index.js` is built and up-to-date before an npm release.)

However, if you feel strongly about including `index.js`, I say go for it. I think @thebearjew and I can handle conflict errors 😄

malgorithms:
ok, sounds good to me.

I was unaware of the `files` attribute in package.json - that's way better and I just switched to that.

I see your point about managing conflicts, so I removed index.js and index.js.map.  I think it's a bit of a bummer that there's no solution here that doesn't slow you 2 down.  I find myself looking in `index.js` for modules a lot, as a quick gander into what's exposed and what the code looks like.

Even with this generated `index.js` it's super fast to see the exported Bot class, and that it exposes `init`, `deinit`, `myInfo`, and `chat`. 

Alternatively, a user would have to:
1. see there's no `index.js`
2. look in `package.json` to find what the entry point is.
3. see that it's `index.js` and be a bit confused
4. realize there's a build step
5. look up what the build script is back in package.json
6. see there's something called `rollup -c` running ("hmm, what is that?" if they've never used rollup)
7. google it or discover by chance there's a `rollup.config.js`
8. note that it has an entry of `lib/entry.js`
9. open that file and see that it just imports and exports Bot
10. finally, go into `lib/bot.js` to discover what they would've seen in `index.js` 

It feels a bit heavy-handed...but like I said I really don't want to slow you guys down, so maybe it's something we come back to later, perhaps once the module is more refined/stable and changing less often. That might be a good compromise. I'll think about this more, but let's leave it the way you like it for now.

PR #51, opened 2018-12-06
Only ignore root index.js for tooling

nathunsmitty:
This was driving me insane for about two hours when I realized prettier wasn't formatting my code.

A couple of index.js files were changed after I linted/prettified/type-checked them. I also version bumped flow and prettier.

### DIFF
--- a/lib/base/index.js
+++ b/lib/base/index.js
@@ -3,7 +3,7 @@ import {keybaseExec} from '../utils'
 import {withStore, actionInit} from '../store'
 import type {Store} from '../store/types'
 import type {DeviceUsernamePair} from '../types'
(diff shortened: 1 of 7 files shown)


lib/base/index.js
```diff
@@ -3,7 +3,7 @@ import {keybaseExec} from '../utils'
 import {withStore, actionInit} from '../store'
 import type {Store} from '../store/types'
 import type {DeviceUsernamePair} from '../types'
-import type {Init, MyInfo, Deinit} from './types'
```
nathunsmitty: All of the type removals was because of eslint's no-unused rule

PR #53, opened 2018-12-07
Downgrade Demos to Promise API

### LINKED ISSUE #45: Downgrade Example Bots from Async/Await
Just to keep things simple with syntax, we're choosing to use the Promise syntax for the example bots rather than async await, to increase legibility for the majority of Node.js readers/writers.

thebearjew:
Also moves promise API to `demo/es7` for the more adventurous.
@malgorithms @nathunsmitty 

Closes #45


demos/math-bot.js
```diff
@@ -0,0 +1,67 @@
...
+            bot.chat.send(message.channel, reply).catch(error => {
+              console.error(error)
+              shutDown()
+            })
+          }
+        }
+      })
+    })
+    .catch(error => {
+      console.log(error)
```
nathunsmitty:
```suggestion
      console.error(error)
```

thebearjew:
es7 directory was @nathunsmitty's idea!

Issue #16, opened 2018-11-08
Chat API Parameter Checking


thebearjew:
When required parameters like `{options: {channel: {...}}` are missing from API methods like `chatSend`, then the library should return a more precise error before invoking the keybase chat api via the service.

thebearjew:
@nathunsmitty thoughts? I do wish that Flow could auto-generate some parameter checking code, but alas.

nathunsmitty:
I support! I suppose an alternative could be passing required options as parameters to the method as opposed to properties in an `options` object. What are your thoughts on that?

Either way, I'm always a fan of nicer error messages!

thebearjew:
Explicit parameters to the api methods would be nicer and easier to check.
Keeping the options object is closer to how the chat api works via the Keybase service on the command line.
Perhaps making this a better library means straying from the CLI api.

nathunsmitty:
My vote is for better library 👍

thebearjew:
This will be handled after #43 lands

PR #56, opened 2018-12-08
Validate Arguments to API Functions

thebearjew:
Instead of allowing missing/malformed types through to the keybase service, catch them in the module and return a more helpful error 🌈 

@nathunsmitty ::insert use typescript here::

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -6,6 +6,7 @@ import crypto from 'crypto'
 import snakeCase from 'lodash.snakecase'
 import {spawn} from 'child_process'
 import readline from 'readline'
+import type {ChatChannel, ChatMessage} from './chat/types'
 
 type ExecOptions = {|
   stdinBuffer?: Buffer | string,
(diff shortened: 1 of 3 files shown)


thebearjew:
Closes #16

lib/utils.js
```diff
@@ -74,3 +75,78 @@ export const formatOptions = (options: any): any =>
...
+  if (!channel.hasOwnProperty('name')) {
+    throw new Error(key(name, 'name', fn))
+  }
+
+  if (!channel.name) {
+    throw new Error(missing(`${name}.name`, fn))
+  }
+}
+
+export const validateChatMessage = (message: ChatMessage, fn: string, name: string = 'message') => {
```
nathunsmitty: would it make sense to have chat validation stuff in the `chat/` dir?

nathunsmitty:
I'm not sure if we need this after #87 lands – I pass error messages from the client into JS errors, so we get errors about arguments from the client as a result.

PR #62, opened 2018-12-11
recursive cleanup + keybaseHomeDir rename

malgorithms:
hi @nathunsmitty and @thebearjew - 

coupla things:

1. `keybaseHomeDir` -> `homeDir` , as requested.
2. I implemented a rmdir that cleans up all the levelDB and other crap that keybase makes. It works, although I wouldn't mind you taking a look at the function to give me feedback on ES6 code style escp. for async/await functions, which is all new to me. Here's the fn; I noticed that `util` coming with node 8+ comes with a promisify, so this all works without external deps. This kind of stuff sure is prettier in iced coffee script :-)

```javascript
export async function rmdirRecursive(dirName: string): Promise<void> {
  const fsLstat = promisify(fs.lstat)
  const fsUnlink = promisify(fs.unlink)
  const fsRmdir = promisify(fs.rmdir)
  const fsReaddir = promisify(fs.readdir)
  try {
    const dirStat = await fsLstat(dirName)
    if (dirStat) {
      for (const entry of await fsReaddir(dirName)) {
        const entryPath = path.join(dirName, entry)
        const stat = await fsLstat(entryPath)
        if (stat.isDirectory()) {
          await rmdirRecursive(entryPath)
        } else {
          await fsUnlink(entryPath)
        }
      }
      await fsRmdir(dirName)
    }
  } catch (error) {
    return error
  }
}
```

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,8 +18,8 @@ type ExecOptions = {|
 // it keeps a service running and acts on stderr, so let's not mess with the logical
 // complexity in keybaseExec
 //
-export function keybaseServiceStartup(keybaseHomeDir: string): Promise<void> {
-  const child = spawn('keybase', ['--home', keybaseHomeDir, 'service'])
+export function keybaseServiceStartup(homeDir: string): Promise<void> {
+  const child = spawn('keybase', ['--home', homeDir, 'service'])
   const lineReader = readline.createInterface({input: child.stderr})
   return new Promise((resolve, reject) => {
     child.on('close', code => {
@@ -36,11 +36,11 @@ export function keybaseServiceStartup(keybaseHomeDir: string): Promise<void> {
 }
 
 export const keybaseExec = (
-  keybaseHomeDir: string,
+  homeDir: string,
   args: string[],
   options: ExecOptions = {stdinBuffer: undefined, onStdOut: undefined}
 ): Promise<any> => {
-  const runArgs: string[] = ['--home', keybaseHomeDir, ...args]
+  const runArgs: string[] = ['--home', homeDir, ...args]
   const child = spawn('keybase', runArgs)
   const stdOutBuffer: Buffer[] = []
   const stdErrBuffer: Buffer[] = []
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -6,6 +6,7 @@ import crypto from 'crypto'
 import snakeCase from 'lodash.snakecase'
 import {spawn} from 'child_process'
 import readline from 'readline'
+import {promisify} from 'util'
 
 type ExecOptions = {|
   stdinBuffer?: Buffer | string,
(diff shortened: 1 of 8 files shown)


lib/utils.js
```diff
@@ -116,11 +117,26 @@ export function mkdir(dirName: string): Promise<void> {
...
+        if (stat.isDirectory()) {
+          await rmdirRecursive(entryPath)
+        } else {
+          await fsUnlink(entryPath)
+        }
+      }
+      await fsRmdir(dirName)
+    }
+  } catch (error) {
+    return error
```
nathunsmitty: I'm not sure we want to return the error here. We should either:

1. Let the error propagate up and handle it in `deinit`.
2. Handle the error, probably via logging to `console.error`.

I think I'm more in favor of handling it in `deinit` and logging that we couldn't delete it directory in there as that keeps `rmdirRecursive` a little more generic in case we need to use it in a future part of the bot and handle errors differently.

lib/utils.js
```diff
@@ -116,11 +117,26 @@ export function mkdir(dirName: string): Promise<void> {
...
 
-export function rmdir(dirName: string): Promise<void> {
-  return new Promise((resolve, reject) => {
-    // The keybase service will have made some files here; we want to clean them all out
-    // Node doesn't have a recursive dir clean, but.
-    console.log(`TODO: implement recurseive rmdir for tmp directory ${dirName}`)
-    resolve()
-  })
+export async function rmdirRecursive(dirName: string): Promise<void> {
+  const fsLstat = promisify(fs.lstat)
```
nathunsmitty: Since this code is mostly synchronous aside from the `await`s, would it make sense to use the synchronous node functions, e.g., `fs.lstatSync`?

lib/utils.js
malgorithms: and ah right, I was basically doing this wrong by returning an error, that would've effectively made the promise resolve instead of reject, but with an error? is that right?

nathunsmitty: yup! The promise would return the error when it resolves and never reject. This also doesn't really make sense because our function signature says we're returning `void`.

lib/utils.js
nathunsmitty: But isn't this how `await fsLstat` works?  The code will block until `fsLstat` resolves or throws an error due to the `await` (I think). 

It seems like the synchronousness of this function is kinda unavoidable: we need to wait on `lstat`  to give us `dirStat` before we can call `readdir` to get `entry` before we can call `stat` on entry, etc.

thebearjew: `await fsLstat` is syntactic sugar for `somePromise.then(result => ...)`

It looks imperative but it isn’t. Other execution can occur (in other promises or functions) until the time when `somePromise` resolves.

nathunsmitty: ^^ just came back from doing a lot of reading about async/await nitty gritty, agree with @thebearjew. We don't want to use the synchronous versions here

thebearjew: 👍 Awesome. It's confusing at first because `await` is only blocking within the execution of a function prefixed with `async`.

I always imagine the code that follows a line with `await` to be magically nested into the body of `somePromise.then(result => { ...  rest of my code ...})`

malgorithms: this is basically what we've been doing with max's iced coffee script. It's nice to see it finally in javascript after all these years. Async code that looks imperative. I'm still figuring out the JS version. In iced you can do nice things such as:

```
for thing in things
   await doSomething thing, defer result
```
which would `doSomething` a bunch of times, and of course be non-blocking. But if you wanted to do them all concurrently you'd just move the await outside:

```
await
  for thing,i in things
    doSomething thing, defer result[i]
```

similarly:

```
await doOnething defer result1
await doSecondThing defer result2
console.log(result1, result2)
```
does them without blocking the JS thread from doing other work, but does them sequentially, and

```
await
  doOnething defer result1
  doSecondThing defer result2
console.log(result1, result2)
```
does them concurrently.

this is all very easy to refactor. I wish it were as clean in Javascript's version...it seems not to be as far as I can tell. I think you've gotta move them into a `Promise.all` call with an array of the promises and then await that, which seems ugly. But I'm still learning, so I could be wrong. In any case this, this is way better than callback hell, and I think better than looking at promises directly which aren't much better. Enjoying learning this.

PR #67, opened 2018-12-12
call service stop

malgorithms:
This is now fixed as of today's admin build of Keybase.


nathunsmitty:
just tested it on my computer, works for me

PR #68, opened 2018-12-12
Coyne/jest tests

malgorithms:
@nathunsmitty and @thebearjew , this introduces `yarn test` which does 2 jest tests so far. Take a look at `__tests__/alice_init.test.js` for the simplest example. Exciting!

Once this is merged I'll fix the process shutdown issue with the watcher subprocess and then we can have tests that make sure deinit() actually kills everything.


malgorithms:
In the interest of keeping things moving while you 2 are at a busy time at school, I'm gonna merge this. This way my subsequent work can have tests and I'm not dealing with a branch of a branch. If there's anything you dislike about the way of set this up, we can discuss it in future issues.

PR #70, opened 2018-12-13
Class Restructure

### LINKED ISSUE #55: Questioning this store/state/reducer model for this module
hi @thebearjew @nathunsmitty 

I've spent a bunch of time with the chatbot code yesterday and today, and I think I'm not sold on the introduction of the `store`. For one, the primary state of Keybase is managed in the actual keybase service itself, so any given Bot only needs to keep track of a few things (its username, working dir, etc.).

But more important, it seems like just for these few things, it's creating a lot of extra code and wrappers with little gain.

As a benefit, it's not like we ever need to rewind state, and I don't think tests will work on loaded or de-serialized state.

As an example of increased effort, I was just working on allowing one-shot mode.  I needed a bot to be aware of its working directory, a value it would generate for itself.  I'd imagine it would simply do something like this in its constructor and somehow save this info so it could be used later:

```javascript
workingDir = utils.randomTempDir()
```

this is easy to follow and only a line of work, exposing this information as needed.

Instead, I had to go into this `store/index.js` and edit the default initialState to add this:

```javacript
keybaseHomeDir: undefined,
```

Then I had to go to bot's init code and add this line:

```javacript
store.dispatch(actionPreInit(update))
```

Then I had to go to `store/actions.js` and create this special type:

```javacript
type PreInitProps = {|
|}
```
 
and an action:

```javacript
export const preInit = (props: PreInitProps): Action => ({
  type: 'PREINIT',
  payload: props,
})
```

and then I needed to go back to `store/index.js` and import and export it:

```javacript
import {actionPreinit} from './actions'
/* then later */
export {actionPreinit}
```

then back in `lib/base.js` I again to import it:

```javacript
import {actionPreinit} from '../store'
```

So I can finally "dispatch" this action to change this one value:

```javacript
    const update = {
      keybaseHomeDir: randomTempDir(),
    }
    store.dispatch(actionPreinit(update))
```

Also need to add this to `action/types.js`:

```javacript
export type PreinitAction = {|
  type: 'PREINIT',
  payload: {|
    keybaseHomeDir: string,
  |},
|}
```

and then redefine action to accept this new type:

```javacript
export type Action = InitAction | DeinitAction | PreinitAction
```

and then `store/types.js` need to handle the update type:

```javacript
+keybaseHomeDir: void | string,
```

Finally need to create a reducer in `store/reducer.js`:

```javacript
    case 'PREINIT': {
      return {
        ...state,
        keybaseHomeDir: action.payload.keybaseHomeDir
      }
    }
```

I'm guessing some of these steps could've been pared down, but it really does seem like even cutting it in half would leave an amazing amount of abstraction a contributor would have to page in.

As a final difficulty, it then leads to a lot of this kind of wrapping, just to get to the state: 

```javacript
const _somethingToDo = (store: Store) => {
  const somethingToDo: whatever = async () => {
    doSomething()
  }
  return somethingToDo
}
```
and then later exporting:

```javacript
somethingToDo: withStore(store)(_somethingToDo),
```

Again, this seems like a lot of function wrapping and returning, when in fact all along a `bot` could simply have access to its data.

I'm not convinced we're getting as much value from this as the effort.  And to be clear, I'm not saying we need to rewind this functional restructure right now, but just letting you know I'm thinking about it. I'll think about it more this weekend.

nathunsmitty:
Keybase bot, now with 100% more object-oriented goodness! Closes #55.

This is a more polished and updated version of my original OO approach (#35) before we decided to take the more functional route. The main idea is that we have a `BaseBot` class that handles initialization stuff and specialized bots that inherit from that bot, e.g., `ChatBot`. Our `Bot` class is also a specialized, but one that is composed of other bots, that it handles the initialization and reinitialization process for.

In order to handle this `Bot` being able to initialize other bots, I ended up having to implement a sort of  "hijack mode" method to `BaseBot`, in which a Bot can be initialized using a currently running service and logged in user's credentials (`initWithExistingService`).

I'll make some comments on the code, but things should hopefully use less code/be clearer to read overall.


lib/base/index.js
```diff
@@ -1,136 +1,138 @@
...
-    !devicename ||
-    currentDPair.username !== username ||
-    currentDPair.devicename !== devicename
-  ) {
-    throw new Error('Uh-oh, username has changed or we never initialized correctly.')
-  }
-}
+export type InitOptions = {|
+  verbose?: boolean,
+  homeDir?: string,
```
nathunsmitty: I think it makes sense to allow users to define their own `homeDir` if they so desire, and randomly generate one if not specified.

lib/base/index.js
```diff
@@ -1,136 +1,138 @@
...
+      // this._log(`intialized ${currentDPair.username} (device=${currentDPair.devicename})`)
+    }
+  }
+
+  /**
+   * Deinitializes the bot by logging out, stopping the keybase service, and removing any leftover login files made by the bot. This should be run before your bot ends.
+   * @memberof BaseBot
+   */
+  async deinit(): Promise<void> {
+    if (this.initialized === 'paperkey') {
```
nathunsmitty: It seems fair to only logout and stop the service if a user `init`ed the standard way. If they use an existing service, I think we want to do nothing on a `deinit`

lib/base/index.js
```diff
@@ -1,136 +1,138 @@
...
+export type DeviceUsernamePair = {|username: string, devicename: string|}
 
-/**
- * @namespace bot
- */
+class BaseBot {
+  initialized: Initialized
+  username: void | string
+  devicename: void | string
+  homeDir: void | string
```
nathunsmitty: `homeDir` can be `void` in the case where someone uses the standard home directory version of the service.

lib/chat/types.js
```diff
@@ -55,7 +55,7 @@ export type ChatOptionsSearch = {|
 
 /**
  */
-export type ChatListOptionals = {|
```
nathunsmitty: I think `Optionals` should be `Options`?

lib/utils/formatOptions.js
```diff
@@ -0,0 +1,19 @@
+// @flow
```
nathunsmitty: `utils` is also starting to get kinda big, so I think a `utils/` folder with each individual function in a file is a little more manageable.

lib/chat/types.js
```diff
@@ -161,11 +161,11 @@ export type ChatOptionsSearchRegexp = ChatOptionsSearchInbox & {|
 /**
  */
 export type AnyChatOption =
```
nathunsmitty: Not sure we need `AnyChatOption` or the Chat API Function Signatures anymore. I can delete but lemme know

thebearjew: Destroy at will

lib/base/index.js
malgorithms: Allowing this may be dangerous with the current code, as the cleanup process expects this to have been generated randomly and have a big random string in it. Deinit right now (1) deletes the directory, and (2) kills any process that _mentions_ the directory. (This second technique isn't totally ideal; I'd like to change it to keep track of spawned pids.) 

In other words, imagine someone passed a homeDir that had shit in it, because they didn't really understand the argument. And then it nuked their files.

So, if you allow passing one, you'll need to make 2 changes:
(1) confirming the directory doesn't already exist (OR if it does only delete known Keybase files/DB's)
(2) change the killing of the listen processes to be smarter than a regex for the homedir.

I guess for now I'd propose dropping this and creating a separate issue for allowing people to specify a home directory.

lib/base/index.js
malgorithms: agreed, with one exception. we still need to kill any processes that were spawned by `watchFor...` calls.

lib/chat/types.js
malgorithms: SGTM

lib/base/index.js
nathunsmitty: Good point. We do need a way right now to give sub bot the current directory of its parent unless we want a bunch of different services running for just one bot, but I'll make that into a private method for now that we can make public in a future PR.

nathunsmitty:
Alright think this should be good for another look over @malgorithms – I made the homeDir argument stuff private for now and also got it to pass your tests.

PR #71, opened 2018-12-13
Coyne/issue 64

malgorithms:
@nathunsmitty @thebearjew , finally have a `deinit` that seems 100% effective. it cleans up the directory and also kills any listen processes.

There's a good test which looks at running processes and the homeDir to make sure that after init and 2 `watch...` calls there is a good directory, and 3 running processes (1 service, 2 listeners) and after deinit makes sure the dir is missing and there are no such processes anymore. 

There's a todo item mentioned in the code which I'll make a separate ticket for, but this seems to be in good shape.

### DIFF
--- a/__tests__/shutdown.test.js
+++ b/__tests__/shutdown.test.js
@@ -32,36 +32,32 @@ async function countProcessesMentioning(substr) {
   }
 }
 
-async function runTest() {
-  it(`alice can init and deinit()`, async () => {
-    await alice.init(config.alice1.username, config.alice1.paperkey)
-    const homeDir = alice.myInfo().homeDir
+it(`alice can init and deinit()`, async () => {
+  await alice.init(config.alice1.username, config.alice1.paperkey)
+  const homeDir = alice.myInfo().homeDir
 
-    // make sure our bot can return a home directory
-    expect(homeDir.indexOf('keybase_bot_')).toBeGreaterThanOrEqual(0)
+  // make sure our bot can return a home directory
+  expect(homeDir.indexOf('keybase_bot_')).toBeGreaterThanOrEqual(0)
 
-    // make sure that homeDir exists
-    expect(await doesFileOrDirectoryExist(homeDir)).toBe(true)
+  // make sure that homeDir exists
+  expect(await doesFileOrDirectoryExist(homeDir)).toBe(true)
 
-    // make sure we see a running server processes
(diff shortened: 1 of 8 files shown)


__tests__/shutdown.test.js
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env node
+/* eslint-env jest */
```
nathunsmitty: We can configure this in `.eslintrc` instead of on a per file basis! I'll make a PR for that.

__tests__/shutdown.test.js
```diff
@@ -0,0 +1,67 @@
...
+
+const fs = require('fs')
+const {promisify} = require('util')
+const Bot = require('../index.js')
```
nathunsmitty: I think we should import `Bot` from `lib` instead of the transpiled `index.js`. Jest code is already going through babel, so it doesn't make a lot of sense to do this twice and require us to run `yarn build` before we can `yarn test`.

Also since the jest code is compiled by babel, we can use es6 imports instead of node's `require` syntax!

__tests__/shutdown.test.js
```diff
@@ -0,0 +1,67 @@
...
+    if (e.code === 1) {
+      return 0
+    } else {
+      throw new Error('Error looking for processes')
+    }
+  }
+}
+
+async function runTest() {
+  it(`alice can init and deinit()`, async () => {
```
nathunsmitty: Is there a reason for wrapping the test in the `runTest` function?

malgorithms: yeah, it's nice to have the test be a series of `await` imperative-looking function calls. but you can't do that except inside an `async` function, I believe. So I had to move it inside a wrapper function like that.

__tests__/shutdown.test.js
malgorithms: cool, I'll change this on all the bots

malgorithms: slight change, rather than `import Bot from '../lib/bot.js'` I did `import Bot from '../lib/entry.js'` because requiring entry matches what the module will be.

__tests__/shutdown.test.js
malgorithms: oh actually, in this case, the async declaration on the fn passed to `it` suffices so it doesn't need a runTest wrapper... I'm indifferent really. one one hand can drop the wrapper, on the other, we might like tests to look alike and some of them may need it.

PR #74, opened 2018-12-14
test for team sends; cleanup other tests

malgorithms:
This adds a new test for sending and reading messge(s) in a team. It's roughly the same as sending in non-team chats, with a slightly different team structure.

The test config file changed slightly, so to get the tests to pass you'll need to edit your local `tests.config.js` file that you make out of `tests.config.example.js`. In particular it now also wants to know about a team.

also edited the alice1 and bob1 DM testing so they can count to 10 together with a unique code, in case there are other concurrent tests.

### DIFF
--- a/__tests__/alice-init.test.js
+++ b/__tests__/alice-init.test.js
@@ -7,8 +7,8 @@ const alice = new Bot()
 
 async function main() {
   it(`alice can init()`, async () => {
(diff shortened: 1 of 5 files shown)


malgorithms:
actually - this is purely test changes and no lib code, and I'm sensitive to your time as a resource, so I'm just gonna merge.

PR #78, opened 2018-12-16
Store spawned children to kill instead of death via regex

nathunsmitty:
Adds a `spawnedProcesses` member variable to our bot, so that it can later kill them in `deinit`. Note that this PR is also based off of #77


lib/base/index.js
```diff
@@ -15,10 +16,12 @@ class BaseBot {
   devicename: void | string
   homeDir: string
   verbose: boolean
+  spawnedProcesses: ChildProcess[]
```
nathunsmitty: I think @malgorithms mentioned somewhere in a previous PR that he wanted to do this by pid – we could also do that, but I don't see any harm in having a record of the `ChildProcess`es themselves.

lib/chat/index.js
```diff
@@ -157,24 +159,21 @@ class ChatBot extends BaseBot {
...
-    keybaseExec(this.homeDir, ['chat', 'api-listen'], {
-      onStdOut: (line: string) => {
-        try {
-          const messageObject = JSON.parse(line)
-          if (this.username && messageObject.msg.sender.username !== this.username.toLowerCase()) {
-            onMessage(messageObject.msg)
-          }
-        } catch (error) {
-          console.error(error)
+    const child = spawn('keybase', ['--home', this.homeDir, 'chat', 'api-listen'])
```
nathunsmitty: I think it makes sense to not use `keybaseExec` for the message watching methods. Unlike most other keybase calls, we don't really care about the return type, we just want our callback to fire on every new line.

Not only does this give us the advantage of being able to add the children easily to our `spawnedProcesses` array, but I think this will make error handling easier for the watch methods, as we don't have to worry about the service erroring on deinit.

lib/base/index.js
malgorithms: oh no agreed, this is probably much better.

PR #80, opened 2018-12-21
Make "use existing service" mode public (Closes #48)

### LINKED ISSUE #48: Use Existing Service Mode
Chris asked for the opposite in #46, but I still want what's currently there. There's a case to be made that the bot should have an initialized level DB with lots of good locally cached data. This is better for *security* and *performance*.  I would personally want to run my bot this way.  So all that's needed is a mode where the existing library can plug into an already running service, in a well-known, non-temporary directory. Thanks!

nathunsmitty:
Takes the previous `_initWithExistingService` that I introduced in #70 and renames it to the slightly clearer and public `initFromRunningService`. Like its name may imply, `initFromRunningService` allows users to run bots off of a running service with a logged in user.

Accompanying this change is two suites of expanded testing surrounding initialization and deinitialization, called `init.test.js` and `deinit.test.js`, which would take the place of `alice-init.test.js` and `shutdown.test.js` respectively. `init.test.js` and `deinit.test.js` are a little more unit test-y than our previous tests and they have a bit better coverage of all possible inputs/documentation on expected behavior. I thought I'd write them as a proposal as to how we could get finer-grained testing while still testing actual usage of the service without mocking. Let me know what you think!


nathunsmitty:
should be ready to go with the new restructure!

nathunsmitty:
Also added some `myInfo` tests and checked the boxes for tests that we fulfilled in #14

nathunsmitty:
@malgorithms bump

PR #81, opened 2018-12-21
restructuring classes some more

malgorithms:
@nathunsmitty and @thebearjew I liked Nathan's recent refactoring of the class system of the bot. I ended up spending a bunch of time over the last few days and ended up realizing one structural change would make sense.

both `chat.js` and the exported `Bot` in index.js were inheriting from a common `BaseBot`, but I realized they were sharing almost no code in common. their inits were different, their deinits were different, their internal state was different, and there was a lot of custom logic or separate functions to manage this.  I  think maybe they'd diverge more as we started adding more functionality to the "service" side of things, and started adding more client functionality than just chat.

So...without a lot of code changes I did some basic restructuring.

* `index.js` exports a Bot class, obvs. That hasn't changed. 

* that Bot can `init` a `Service` instance if it needs to, or it can talk to an existing one. (Coming soon.) The Service class doesn't inherit from a common base because it's just different. But a `Service` knows to shut down itself, remove its directory, etc. when being deinitialized.

* The Bot has a "ChatClient" inside it, exposed as `this.chat`. It inherits from a common class called `ClientBase`; when we want to add say `bot.wallet` or `bot.team` or whatever, those will all be `ClientBase` objects too. they'll share a lot of code such as how to make API function calls, how to guard initialization, etc.

This simplifies the top level bot. For example its init right now looks like this:

```
    await this._service.init(username, paperkey, options)
    await this.chat._init(this.myInfo().homeDir, options)
```

And its deinit looks like this:

```
    await this._service.deinit()
    await this.chat.deinit()
```

open to feedback. I think it's cleaner to read and follow, and it seems like it'll be less work to add stuff because BaseBot won't be doing 2 different kinds of things.


lib/client-base/index.js
```diff
@@ -0,0 +1,86 @@
...
+    this.homeDir = homeDir
+    this.username = initBotInfo.username
+    this.devicename = initBotInfo.devicename
+  }
+
+  /**
+   * Deinitializes the client by killing any spawned processes
+   * @memberof ClientBase
+   */
+  async deinit(): Promise<void> {
```
nathunsmitty: I think we should add an `_` before deinit and hide it in the docs, we don't want users to be able to call `bot.chat.deint()`
```suggestion
   async _deinit(): Promise<void> {
```

malgorithms: oh, good call.

PR #87, opened 2018-12-31
[WIP] Comprehensive chat tests

nathunsmitty:
Adds a `chat.test.js` file that has comprehensive tests for all chat methods. @thebearjew's suggestions in #14 was a good starting point, and I added a couple tests from there.

A couple of notes on things I found during the making of this PR:
- I ended up tightening up a lot of flow types and error handling!
- Although we were formatting input objects to the service (e.g., `{unreadOnly: true}` -> `{unread_only: true}`) we weren't formatting output from the service back into camelCase. I generalized our `formatOptions` function into two functions for input and output. (And also added tests for those!) As a result, `lodash.camelCase` was added as a dep.
- While our flow types for chat objects are very useful for documentation, I think it's important to know that because we're `JSON.parse` (which has a return type of `any`), we aren't guaranteed type safe return types. It's important to check these at runtime with Jest, which is what I tried to do with the "matcher" objects in the suite. They could probably be even stricter, though.
- Jest will parallelize test files, which means we can encounter race conditions when doing integration tests like these. However, Jest will [run test files sequentially](https://jestjs.io/docs/en/setup-teardown.html#order-of-execution-of-describe-and-test-blocks) so I suppose that means we need to put any potential tests that could cause a race condition in the same file, resulting in deterministic behavior at the cost of speed. I think I'm still trying to figure out the right balance between the two.


__tests__/chat.test.js
```diff
@@ -0,0 +1,263 @@
...
+      })
+      const {id} = messages[0]
+      expect(alice.chat.delete(invalidChannel, id)).rejects.toThrowError()
+    })
+
+    it('Throws an error if given an invalid id', async () => {
+      expect(alice.chat.send(channel, -1)).rejects.toThrowError()
+    })
+
+    it('Throws an error if it cannot delete the message (e.g., someone else wrote it)', async () => {
```
nathunsmitty: This error currently fails; I wrote it before I realized that anyone can actually delete any message in a conversation. Just wanted to check before I remove it as I know that feature isn't in the GUI.

malgorithms:
> I noticed that unsupported arguments in chat method's options appear to have no effect on output, i.e., bot.chat.list({fakeOption: 'blah'}) acts the same as bot.chat.list(). This is true at the command line level too: keybase chat api -m '{"method": "list", "params": {"options": {"fakeOption": 'blah'}}}' gives the same result as keybase chat api -m '{"method": "list"}. Not sure if we want to throw an error in JS or preserve the CLI behavior.

I would say we should follow the CLI behavior; extra params passed to something that takes JSON isn't a big deal. It does allow for easier deploy strategies, as a bonus. It's nice when party A and B communicate and some new optional parameter can be added/understood on A's and B's sides in either order, without a coordinated simultaneous restart.

__tests__/chat.test.js
malgorithms: hmm, it was my understanding that in team chats, admins could delete from anyone. and in DM's people could only delete their own. But this is a one-on-one chat, outside a team, and alice is deleting bob's message?

lib/chat-client/index.js
```diff
@@ -29,29 +29,41 @@ class Chat extends ClientBase {
...
+    }
+    return res.conversations || []
   }
 
   /**
    * Reads the messages in a channel. You can read with or without marking as read.
    * @memberof Chat
    * @param channel - The chat channel to read messages in.
    * @param options - An object of options that can be passed to the method.
+   * @returns - A summary of data about a message, including who send it, when, the content of the message, etc. If there are no messages in your channel, then an error is thrown.
```
malgorithms: shouldn't an empty chat just reply with an empty array? instead of throwing an error? it seems to me asking for a list of something and that list being empty doesn't imply an error, just a successful empty list.

nathunsmitty: That was what the client was doing, so I just passed that behavior on. All for making it an empty array though, that makes more sense.

__tests__/chat.test.js
malgorithms: I'll make a core ticket for this.

lib/chat-client/index.js
malgorithms: ok, I think I know why this is happening. let's leave it in core's hands too.

PR #90, opened 2019-01-02
support logging to file

malgorithms:
with this new version the bot keeps a log file in the expected place in its homeDir. This has the added advantage that you can do this if there's a bug:

`keybase --home=(whatever) log send` and get a real service log.

this log is deleted along with everything else when the bot deinits.


malgorithms:
cool, just made small additional changes:

* `lodash.camelcase` was missing from rollup so `yarn build` wasn't working. fixed that.
* `we had 2 promisified timeout functions, one in test-utils and one in utils, so I dropped the one in test-utils.
* commented out your test about alice deleting a message by bob that fails; it's acting as expected for now, so our bot might as well not fail the test

Issue #65, opened 2018-12-12
option or metainfo to include/exclude own messages


malgorithms:
some people like to write bots that act as themselves in a team setting, say so they can ask for reminders or order lunch without adding a bot account to their team, and currently `watchAllChannelsForNewMessages` and `watchChannelForNewMessages` don't include those messages.

Two options:

(1) include a user's own messages but expose some extra meta information on messages, so a bot author can easily filter them out:
  * `sentByMe` - whether the message sender was the bot's user (e.g. bot running as `chris`, sender was `chris)
  * `sentFromThisDevice` - whether the message was sent by the same paperkey/device as the bot is using
 * `sentFromThisBot` - whether it actually was initiated from this exact bot instance. note this is different from `sentFromThisDevice` because you might run 2 totally different bot scripts with the same paperkey, and want them to see each other's messages but not get triggers on their own.

(2) instead of that info coming back in messages, make them filter options on the call. This is slightly more convenient for shorter scripts but has the disadvantage that it complicates the creation of the function, and I'm hesitant to make these functions even more complicated.

This strategy will come up elsewhere. People often only want to watch for text messages or attachments or whatever, or maybe they actually care to get message edits. Currently we're letting the user filter, which seems simpler.  That might be the right answer.

Whichever we choose should get a test.

nathunsmitty:
Just wanted to toss in that if we go with approach 1, then our message types start to deviate from the types that are returned by the service. That being said, we may want to do that as it would make for a better API but I think it's important that we're aware of such a change.

PR #91, opened 2019-01-03
watching for messages now includes ones own from other devices

malgorithms:
@nathunsmitty, small changes here although lots diffs in chat.test.js because the bot `alice` is now `alice1` and `alice2`.

the key change here is that watching for new messages now by default includes your own message sent from other devices. I decided to resolve the issue (#65) in the simplest imaginable way: just let the user look at the data themselves if they want to filter these out. 

so, watching for new messages:
- includes messages from others
- includes messages from your other devices 
- does not include messages from the same device that's running the bot

also note I call your `formatAPIObjectOutput` on the watched messages, which we weren't doing.

### DIFF
--- a/__tests__/chat.test.js
+++ b/__tests__/chat.test.js
@@ -249,12 +249,14 @@ describe('Chat Methods', () => {
       const stopAt = 10
       const convoCode = crypto.randomBytes(8).toString('hex')
       const directChannel = {name: `${bot1.myInfo().username},${bot2.myInfo().username}`}
+      let totalMessagesSeen = 0
       let highestReached = 0
       const onMessageForBot = bot => {
         const onMessage = async message => {
           if (message.content.type === 'text') {
             const body = message.content.text.body
             if (body.indexOf(convoCode) !== -1) {
+              totalMessagesSeen++
               const num = parseInt(body.replace(convoCode, '').trim())
               highestReached = Math.max(num, highestReached)
               if (num < stopAt) {
@@ -271,15 +273,13 @@ describe('Chat Methods', () => {
       const message = {body: `${convoCode} 1`}
       await bot1.chat.send(directChannel, message)
 
-      // we should exit this as fast as possible, but wait up to 10 seconds
-      let i = 0
-      while (i++ < 100) {
+      while (true) {
         await timeout(100)
         if (highestReached === stopAt) {
           break
         }
       }
-      expect(highestReached).toBe(stopAt)
+      expect(totalMessagesSeen).toBe(stopAt)
     }
 
     it('can have 2 users count together', async () => testTwoBotsCounting(alice1, bob))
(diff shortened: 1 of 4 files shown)


__tests__/chat.test.js
```diff
@@ -219,46 +222,67 @@ describe('Chat Methods', () => {
...
+      await alice1.chat.send(teamChannel, {body: 'hello bob'})
+      await bob.chat.send(teamChannel, {body: 'hello alice1'})
 
       await timeout(3000)
       expect(ALICE_IS_SATISFIED).toBe(true)
       expect(BOB_IS_SATISFIED).toBe(true)
     })
+
+    it("Doesn't pick up its own messages from the same device", async () => {
+      const messageText = 'Ever thus to deadbeats, Lebowski'
```
nathunsmitty: 😆

lib/chat-client/index.js
```diff
@@ -196,14 +198,17 @@ class Chat extends ClientBase {
     const lineReaderStdout = readline.createInterface({input: child.stdout})
     const onLine = (line: string) => {
       try {
-        const messageObject: MessageNotification = JSON.parse(line)
+        const messageObject: MessageNotification = formatAPIObjectOutput(JSON.parse(line))
```
nathunsmitty: glad we changed this! thanks for catching that 🎉

__tests__/chat.test.js
```diff
@@ -219,46 +222,67 @@ describe('Chat Methods', () => {
...
-          if (message.content.type === 'text') {
-            const body = message.content.text.body
-            if (body.indexOf(CONVO_CODE) !== -1) {
-              const num = parseInt(body.replace(CONVO_CODE, '').trim())
-              HIGHEST_REACHED = Math.max(num, HIGHEST_REACHED)
-              if (num < STOP_AT) {
-                const reply = {body: `${CONVO_CODE} ${num + 1}`}
-                await bot.chat.send(message.channel, reply)
-              }
+  const testTwoBotsCounting = async (bot1, bot2) => {
```
nathunsmitty: I think it makes more sense to put this in the describe block for `watchAllChannelsForNewMessages`

malgorithms: cool yeah, moving it in there.

malgorithms:
ok @nathunsmitty  - ran `yarn docs`. also changed the test. I think it was just timing out on your machine, in the 10 second internal timeout that test had. I just removed the 10 seconds from it, which seemed unnecessary since jest is doing a timeout on the whole thing anyway. 

interestingly, I couldn't do this because it wouldn't pass eslint:

```
      while (highestReached < stopAt) {
        await timeout(100)
      }
```

because the linter assumes it's impossible for that to ever be satisfied.  I guess it doesn't understand awaits well enough...anyway, I changed it to a slight variation.

all tests pass for me now with the exception of deinit which fails once in a while, but as you said that's an issue in master.

also moved the testTwoBotsCounting into the describe block as you requested.

PR #95, opened 2019-01-03
test: Add a timeout to a deinit test to make it pass

### LINKED ISSUE #94: Fix deinit test
Currently on `master` ([074daf10dd685911d95a2b739b84587a58b1b68b](https://github.com/keybase/keybase-bot/commit/074daf10dd685911d95a2b739b84587a58b1b68b)):

```
› yarn test
yarn run v1.12.3
$ jest
 PASS  lib/utils/formatAPIObject.test.js
 PASS  __tests__/myInfo.test.js
 PASS  __tests__/init.test.js (12.84s)
 FAIL  __tests__/deinit.test.js (14.542s)
  ● Keybase bot deinitialization › kills all spawned processes it creates

    expect(received).toBe(expected) // Object.is equality

    Expected: 0
    Received: 1

      70 | 
      71 |     // All processes should be shut down for alice
    > 72 |     expect(await countProcessesMentioning(aliceHomeDir)).toBe(0)
         |                                                          ^
      73 |     // The service should still be running for bob, as that was started before he was initialized
      74 |     expect(await countProcessesMentioning(bobHomeDir)).toBe(1)
      75 |     await stopServiceManually(bobHomeDir)

      at Object.toBe (__tests__/deinit.test.js:72:58)

 PASS  __tests__/chat.test.js (42.947s)

Test Suites: 1 failed, 4 passed, 5 total
Tests:       1 failed, 29 passed, 30 total
Snapshots:   0 total
Time:        43.399s, estimated 53s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

nathunsmitty:
closes #94

### DIFF
--- a/__tests__/deinit.test.js
+++ b/__tests__/deinit.test.js
@@ -68,6 +68,9 @@ describe('Keybase bot deinitialization', () => {
     await alice.deinit()
     await bob.deinit()
 
+    // Also give a couple seconds for the processes to fully deinit
+    await timeout(3000)
+
     // All processes should be shut down for alice
     expect(await countProcessesMentioning(aliceHomeDir)).toBe(0)
     // The service should still be running for bob, as that was started before he was initialized


__tests__/deinit.test.js
```diff
@@ -68,6 +68,9 @@ describe('Keybase bot deinitialization', () => {
     await alice.deinit()
     await bob.deinit()
 
+    // Also give a couple seconds for the processes to fully deinit
+    await timeout(3000)
```
malgorithms: I was thinking about this with one of my other tests, and I think the problem with picking a specific timeout amount is that it risks being too low (and you get occasional failures/flakes) or too high (and you're waiting an extra few seconds on every test.  Given that there is a global timeout, it seems like you want the success case to exit as fast as possible, in which case I think I prefer this kind of thing:

```
while (await countProcessesMentioning(aliceHomeDir) > 0) {
  await timeout(100)
}
```

either way though this is fine with me.

nathunsmitty: That's a good point – I'm gonna think about this more, I think we could maybe extract the about logic out into a function, e.g., `waitUntilThisConditionIsSatisfied(...)`.

PR #97, opened 2019-01-05
feat(chat): Reactions to chat messages

nathunsmitty:
Note this isn't totally complete yet, I have a couple of questions/potential core issues.

I also added a bot, "flippy", which duplicates the behavior of that keybase coin flip lunch bot. Give it a try!


__tests__/chat.test.js
```diff
@@ -153,6 +151,30 @@ describe('Chat Methods', () => {
...
+
+      await bob.chat.react(channel, messageToReactTo.id, ':poop:')
+      messages = await alice1.chat.read(channel, {peek: true})
+      const reaction = messages[0]
+      expect(reaction.id).toBe(messageToReactTo.id + 1)
+      expect(reaction.content.type).toBe('reaction')
+      expect(messages[1].reactions.reactions.poop).toHaveProperty(config.bots.bob1.username)
+    })
+
+    // it('Throws an error if given an invalid emoji', async () => {
```
nathunsmitty: So it turns out you can give pretty much anything to the chat react method and it seems to work. This looks a little weird on the GUI when it isn't an emoji though:
screen shot 2019-01-05 at 2 15 08 pm
Is this something we should check for in the bot or the client?

__tests__/chat.test.js
```diff
@@ -153,6 +151,30 @@ describe('Chat Methods', () => {
...
+    //   await alice1.chat.send(channel, message)
+    //   const messages = await alice1.chat.read(channel, {peek: true})
+    //   const messageToReactTo = messages[0]
+
+    //   expect(bob.chat.react(channel, messageToReactTo.id, 'blah')).rejects.toThrowError()
+    // })
+    // it("Throws an error if it cannot react to a message (e.g., it's not a reactable message type")
```
nathunsmitty: I couldn't find any message types that I couldn't react too. I could react to delete messages and reaction messages, both of which don't show up in the GUI at all so I'm not sure if this is an intentional feature.

__tests__/chat.test.js
malgorithms: yeah it's a known bug/feature/whatever. someone jokingly used it to make a voting thing, like you could put "Yes" and "No" as 2 reactions and people could pick by clicking. it should be blocked in CORE not by our module, so you shouldn't have to do anything here.

__tests__/chat.test.js
malgorithms: not sure either. I would say if it doesn't cause probs don't worry abbout it

PR #99, opened 2019-01-13
feat(chat): Support sending and downloading attachments

nathunsmitty:
Adds two new methods, `attach` and `download`, which do what you'd expect. Also adds some tests and documentation around the new methods.

### DIFF
--- a/__tests__/chat.test.js
+++ b/__tests__/chat.test.js
@@ -3,6 +3,7 @@ import fs from 'fs'
 import Bot from '../lib'
 import config from './tests.config.js'
 import {timeout} from '../lib/utils'
+import {promisify} from 'util'
 
 test('Chat methods with an uninitialized bot', () => {
   const alice1 = new Bot()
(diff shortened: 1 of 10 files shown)


__tests__/chat.test.js
```diff
@@ -175,6 +176,60 @@ describe('Chat Methods', () => {
     // it("Throws an error if it cannot react to a message (e.g., it's not a reactable message type")
   })
 
+  describe('Chat attach', () => {
+    const attachmentLocation = '/tmp/kb-attachment.txt'
+    beforeAll(() => {
+      fs.writeFileSync(attachmentLocation, 'This is a test file!')
```
malgorithms: I think it'd be better to include `import {promisify} from 'util'` at the top and have all your fs operations async. I don't think blocking the event loop is a good idea. 

something like this, I believe, will get you clean lines without callback hell.

```  
beforeAll(async () => {
    await promisify(fs.writeFile)(attachmentLocation, 'This is a test file!')
  })
```

__tests__/chat.test.js
```diff
@@ -175,6 +176,60 @@ describe('Chat Methods', () => {
...
+      await alice1.chat.attach(channel, attachmentLocation)
+
+      // Read the file
+      const messages = await alice1.chat.read(channel)
+      await alice1.chat.download(channel, messages[0].id, downloadLocation)
+      const downloadContents = fs.readFileSync(downloadLocation)
+      expect(downloadContents.toString()).toBe(attachmentContent)
+
+      // Delete the created files
+      fs.unlinkSync(attachmentLocation)
```
malgorithms: similarly, 

```
 await promisify(fs.unlink)(attachmentLocation)
```

you get the idea

PR #106, opened 2019-01-24
Fixed the startup process on slower systems

pzduniak:
The code only worked before because it ran on systems with fast IO ops. On WSL, where the IO performance is sub-par compared to native code, it resulted in the incorrectly written startup checks succeeding before the service was running. Additionally, the checks themselves were incorrectly written as `-1` is truthy in JS.


package.json
```diff
@@ -53,7 +53,8 @@
   "dependencies": {
     "lodash.camelcase": "4.3.0",
     "lodash.kebabcase": "4.1.1",
-    "lodash.snakecase": "4.1.1"
+    "lodash.snakecase": "4.1.1",
+    "tail": "^2.0.2"
```
heronhaye:
```suggestion
    "tail": "2.0.2"
```

heronhaye: I'm not really sure how yarn.lock works tbh so maybe this isn't necessary

pzduniak:
That's what I thought @malgorithms, which is why I evaluated around 5 libraries that implement this functionality and this one was most widely used, had the simplest implementation (as you said, I did in fact read the npm-published code) and didn't use any additional dependencies. If you would like me to rework this as an util in this repo, I can do it.

Here's the one file that the npm version of `tail` contains: 
https://gist.github.com/pzduniak/de2e5b366f1b7720b2f59d259a86d562

That being said I totally forgot about checking the `package.json`, so thanks Surya!

malgorithms:
@heronhaye I *think* what happens with the `^` version numbers and the yarn lock is that a subsequent `yarn` call inside this module would get the same version of `tail`. However, someone doing a `yarn add keybase-bot` itself wouldn't necessarily be locked in at the same number. I could be wrong, but I think that's how nojima described it to me. In any case best to play it safe with just locking in a version number.

@pzduniak that code looks pretty simple and clean to me, and I'd be fine approving it. Maybe you can get one more eye on it from the keybase team?  That's not new code, right? It's been live in the npm registry for a month, right?

heronhaye:
@malgorithms so should there be a yarn.lock diff here then?

malgorithms:
yeah, that's my understanding

malgorithms:
@pzduniak - did you add the package with `yarn add tail` or just manually add it to `package.json`?

pzduniak:
Ah yes, sorry, I haven't noticed that we're using `yarn`, I personally migrated back to npm after they added lockfiles so I'm not used to it. Change pushed.

malgorithms:
ah cool. 

@pzduniak, can you also do a `yarn flow` and `yarn test` to make sure all that stuff passes?

pzduniak:
`yarn flow` is all OK, for `yarn test` I can't run since I'd need to configure them, not sure if it'd be worth spending time on creating new accounts for that purpose if most likely I won't have to deal with modifying this library again for my assigned project...

That being said there's no reason why it wouldn't work, I manually tested it out, using some of the more complicated features and it's been working perfectly ever since I fixed this thing.

malgorithms:
a couple things:

- tail needs to be added to `rollup.config.js` as an external requirement. `yarn build` constructs the `index.js` file that we publish to npm, and it needs to understand that it's an external requirement. It's just a 1-line fix and should be clear from how we handle lodash there.

- this isn't passing tests. something is up in initialization. I don't have time to look into it right now, but in particular if you set up a test config file and try even just the init test, some of them will fail:

`yarn test`

if you want to just try a small set of tests, do:

> yarn test __tests__/init.test.js

```
  Keybase bot initialization
    ✕ can init with a username and paperkey (60011ms)
    ✓ throws an error if not given a username (3ms)
    ✓ throws an error if not given a paperkey
    ✓ throws an error if given an invalid paperkey (914ms)
    ✓ can init if a service with a logged in user is currently running (1428ms)
    ✓ throws an error if an invalid home directory is given as the location of a currently running service (46ms)
    ✕ throws an error if it is already initialized (60004ms)
```

this is on my mac, if that matters.

pzduniak:
Ran the tests for 4 times on both WSL/Linux and OSX - everything worked. Then each test started having a ~25% chance of randomly failing. I have no clue what's up with that but I'm looking into this.

pzduniak:
I think I've fixed the issue on WSL, there was a race condition on service shutdown vs cleanup, the darwin thing seems to be completely different though, investigating now.

pzduniak:
Looks like that I fixed both race conditions, please run the tests locally a few times to make sure that it's not occurring. Couldn't reproduce it anymore on both WSL and OSX on my most recent commit.

PR #107, opened 2019-01-24
bot lite mode

joshblum:
`-enable-mode-lite-mode` supersedes `KEYBASE_DISABLE_TEAM_AUDITOR` and disables some other non essentials. depends on https://github.com/keybase/client/pull/15672 merging


lib/service/index.js
```diff
@@ -12,6 +12,7 @@ import {promisify} from 'util'
  */
 export type InitOptions = {|
   verbose?: boolean,
+  botLite?: boolean,
```
malgorithms: would be nice to add some notes above in the comment about what `botLite` is

PR #116, opened 2019-01-31
Create / join / leave channel support + read pagination

pzduniak:
Things to consider:
1) How would we test it? Team support seems to be required for this
2) Is it OK to have thousands of channels in a single team, as it will inevitably happen?

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,33 +97,48 @@ This code is also in [`demos/hello-world.js`](demos/hello-world.js), if you want
   - [BotInfo](#botinfo)
     - [Properties](#properties-1)
 - [Chat](#chat)
-  - [list](#list)
+  - [join](#join)
     - [Parameters](#parameters-2)
     - [Examples](#examples-4)
-  - [read](#read)
+  - [leave](#leave)
     - [Parameters](#parameters-3)
     - [Examples](#examples-5)
-  - [send](#send)
+  - [list](#list)
     - [Parameters](#parameters-4)
     - [Examples](#examples-6)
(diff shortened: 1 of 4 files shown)


malgorithms:
good q's. One idea might just be a nice full test that does something like these in order:

1. create a random channel
2. list channels and make sure it's there
3. join the channel 
4. list channel's members and make sure we're in there
5. leave the channel
6. list channel's members and make sure we're out of it
7. try to post in the channel and confirm it fails in the way we want (now that we're out of it)
8. delete the random channel
9. list channels and make sure it's gone

this could all be done in the acme team that is declared in the config, so we wouldn't need team creation stuff. But it would need to be able to do all of the above in order. What do you think?

pzduniak:
I actually had no idea that we could delete channels! It's so hidden in the UI (though I guess this is desirable, I wonder how it works internally).

I'm not 100% sure what's the difference between teams and big teams, I guess it's just about the channels count? It looks like we're just missing the channel creation, which would solve #24.

pzduniak:
This is tough, I tried to add deleteConv to the API, but it requires UI or at least some sort of a UI emulator - somehow I also wasn't able to script this out and after a short convo with core we decided to do it next sprint. Did the second best thing, used a static secondary convo to make this work - effectively we're testing out _calling_ the endpoint rather than the endpoint itself, so I think it should be sufficient for this lib.

pzduniak:
What about `join` and `leave`? I left them as-is, is that okay?

PR #118, opened 2019-02-04
Pagination.last should be a boolean.

Bannerets:
https://github.com/keybase/client/blob/2e662bdec0d7f0c1e4f49235cb0ae13b47050480/protocol/avdl/chat1/common.avdl#L157

With string it shows `001 json: cannot unmarshal string into Go struct field Pagination.last of type bool` error.

### DIFF
--- a/lib/chat-client/types.js
+++ b/lib/chat-client/types.js
@@ -51,7 +51,7 @@ export type Pagination = {|
   num: number,
   next?: string,
   previous?: string,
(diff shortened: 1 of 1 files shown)


pzduniak:
Thank you for your contribution, this indeed seems to be the case, although from what I have understood you don't necessarily want to pass "last" in the options object to read. Nevertheless, good catch.

PR #121, opened 2019-02-05
Include index.js in the repo

pzduniak:
useful for pinning the lib in package.json to a branch


pzduniak:
ack'd by @malgorithms in a dm

PR #123, opened 2019-02-08
Fix EditContent in flow typings: 'messageID' -> 'messageId'

Bannerets:
Strangely here https://github.com/keybase/client/blob/b6aadbbd398c09010a6469de78f68b5db80cc9c6/protocol/json/chat1/local.json#L126 it's also written as `messageID`, but I receive `messageId`:

image

I guess `formatAPIObjectOutput` function changes that?

**upd**: Yep: `require('lodash.camelcase')('messageID') //=> 'messageId'`
 Also it doesn't touch `messageIDs`: `require('lodash.camelcase')('messageIDs') //=> 'messageIDs'`

### DIFF
--- a/lib/chat-client/types.js
+++ b/lib/chat-client/types.js
@@ -122,7 +122,7 @@ export type ReactionContent = {|
 export type EditContent = {|
   type: 'edit',
   edit: {|
-    messageID: number,
+    messageId: number,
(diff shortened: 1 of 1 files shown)


pzduniak:
Thanks!

PR #124, opened 2019-02-16
add msg id from response to send/attach/react

songgao:


songgao:
I did not! Thanks for brining it up. It doesn't seems like it produces new top level item, but does produces some changes in the doc. In addition to the return type change, it also seem to have shifted stuff around (e.g. adding spaces). Perhaps some version mismatch for a package in node_modules?

Also, do you plan to make a new npm release soon?

pzduniak:
The spaces thing is fixed by prettier, it's just a weird thing that gets sorted out by the pre-commit hook - I have no clue why `documentation` does that. I feel that we should eventually make the doc gen print out all the result types in the README too, but this is not urgent so noone _really_ cares about it that much. wrt release ask @malgorithms

songgao:
Cool thanks Piotr; merging!

PR #126, opened 2019-02-18
Coyne/fix watch channel bug

malgorithms:
@pzduniak review?

this:

1. fixes the fact that we were calling params `topic_name` and `topic_type`, while we used camelCase everywhere else

2. fixes the bug @songgao pointed out.

3. comments out the test that's failing for me.

4. has a version bump to 2.0.0. since the variable changes, and since your pagination changes.


malgorithms:
@songgao if you want to give me a thumb on this today I can publish 2.0.0.

lib/chat-client/index.js
```diff
@@ -383,7 +383,7 @@ class Chat extends ClientBase {
         } else if (
           // fire onMessage if it was from a different sender or at least a different device
           // from this sender. Bots can filter out their own messages from other devices.
-          (!channel || channel.name === messageObject.msg.channel.name) &&
+          (!channel || channel.topicName === messageObject.msg.channel.topicName) &&
```
songgao: I think we still need to check `name` here, and other fields in `channel`. Perhaps have a helper function [like this](https://github.com/keybase/kirabot/blob/fa96e8d323db85cc674bd1803b115ae50a715769/src/kira.js#L50-L54)?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "keybase-bot",
-  "version": "1.0.4",
+  "version": "2.0.0",
```
songgao: 🎉

__tests__/chat.test.js
```diff
@@ -185,15 +187,23 @@ describe('Chat Methods', () => {
...
+        topicType: 'chat',
+        membersType: 'team',
+        topicName: 'subchannel',
+      }
+      const generalChannel = {
+        name: config.teams.acme.teamname,
+        public: false,
+        topicType: 'chat',
+        membersType: 'team',
+        topicName: 'general',
```
songgao: It'd be nice to have a few different types of channel here, like different team, topic type, membersType, etc.

__tests__/chat.test.js
```diff
@@ -206,6 +216,18 @@ describe('Chat Methods', () => {
       expect(read1.messages[0].content.type).toEqual('join')
       expect(read1.messages[0].sender.username).toEqual(config.bots.bob1.username)
 
+      let bobMessageCount = 0
+      const bobOnMessage = async message => bobMessageCount++
+      bob.chat.watchChannelForNewMessages(teamChannel, bobOnMessage)
+      bob.chat.watchChannelForNewMessages(generalChannel, bobOnMessage)
+      await timeout(500)
+      await alice1.chat.send(generalChannel, message)
+      await timeout(500)
```
songgao: Do we wanna check the received message is from the specified channel? Same below. Doesn't seem absolutely necessary though.

lib/chat-client/index.js
malgorithms: good catch on the name.  It would need to be a bit more complicated than your helper function, because it would need to understand the defaults. For example, the `public` field and some of the other fields are optional, and users may not pass these in. 

But yeah, I'll add this, thanks for catching.

songgao: Oh good point! WDYT about exporting that helper function so library users can use?

Also, not suggesting such a change in this PR but for future I think this would be solved by `immutable.js` where we populate default values when creating objects just like in FE code.

songgao: `immutable.js` brings some other benefit though. For example, when comparing channels we can do a simple reference comparison with `===` and if it's the same we know it's the same channel, and if not we'd compare the fields.

But yeah, it'd be a big opinionated dependency and if we do that we'd probably want to do it for other types too which would be a big change. Not sure; just thinking out loud!

malgorithms: @pzduniak agreed about spawning separate processes. Although, on the other hand, I dislike how the bot needs to process everything that comes through keybase chat. If you just want to watch one channel, it stinks your bot needs to get everything from every team and ever channel and filter it out. I almost wish the chat api listen call took a channel as an argument. (But that would keep us on this multi-process model.)

Song, I think I dislike bringing immutable in because it's, as you say, massive, and we're trying to keep this lightweight and simple. `===` won't work anyway. These channel comparisons don't actually match. A user asks to listen to {name:"keybase",topicType:"general"} and what comes back is {name:"keybase",public:false,topicType:"general",membersType:"team", // etc. So somewhere in this code we need to understand the idea of defaults and map their abbreviated request to a full options object. If we do that, then we can use `===` but then we're just going through all the steps of comparing anyway. 

anyway, lame of me to try to race this out the door and break the team check. I'll get this fixed today before merging.

__tests__/chat.test.js
malgorithms: we do want to bring back concurrency tests. I think the biggest issue with them is that there are some tests that use the same actual channels. For example, I think there's an inline chat send of some money from alice to bob and then alice or bob looks at the most recent message to make sure it has that info. The tests will have to change a bit.

malgorithms:
@songgao and @pzduniak we had a problem where our new channel matching wasn't just missing the team name check and letting the wrong things through.. It was _also_ doing the opposite, and filtering out messages that were valid but where the channel name was provided in reverse order (`chris,max` vs. `max,chris`) or excluding the watcher in the name. (For example, before this fix I could send to the channel `{name: 'max'}` and then listen to that same channel `{name:max}`, and the former message would go through, but the latter wouldn't actually get the message....because it's missing the `,chris`. 

I added some normalization of the channel name and also added some tests to fix this stuff.

All this normalization stuff really points to me at the idea that a client library shouldn't have to do this channel understanding; ideally you subscribe to a channel by telling the keybase process to send you messages from that channel, and then you only actually get messages from that channel. I've made a separate ticket for that. It will require some core changes, especially if we want to do it all in one process with multiple such subscriptions.

As song mentioned, we should also add some tests for other params, for example posting where public = true or topic type = 'dev'. But I don't think those messages actually come back from chat api listen (not sure)... that can be a separate ticket too. It should come after our ideal solution for rearchitecting listen.

songgao:
Ah good point; I missed that in my review.

In GUI it's usually referenced by `ConversationID`. Can we have something that lets us resolve a `ChatChannel` into a `ConversationID`, and use that for managing the subscriptions?

pzduniak:
LGTM now! wrt the matching issues: wow, I wonder how come I never ran into this issue with the compliance stuff. This should definitely be handled by the API itself (and I assumed that it was), we already have plenty of code handling weird incompatibilities already.

malgorithms:
Hmm yeah, `ConversationID` could be fine for referring to a channel object. It's not something that should ever be required for a user of the library, even if used by the library internally. And it shouldn't require something weird like the user "creating" the convo first. For example, I should be able to start listening to `max,chris` before either of us has ever written into it. not sure how that works...placeholder conversationID's ? in any case we can discuss that in the other ticket.

PR #132, opened 2019-03-14
Coyne/channel watching service side

malgorithms:
@songgao  review? previously the node module would have to use chat api listen to get _all_ incoming messages, even when the bot called watchChannelForNewMessages. And then the module would filter out anything that didn't match the channel. This was extra node processing work and error prone, since channels can be referred to in lots of different ways. So now the filtering is passed as a command line option.

The test changes here are just some cleanup, the only code change is in chat-client/index.js. Note how it's almost all code removal! yay.


__tests__/chat.test.js
```diff
@@ -228,10 +229,16 @@ describe('Chat Methods', () => {
...
+      const allChecksOk = () => {
+        for (const i in channels) {
+          if (!okChecks[i]) {
+            return false
+          }
+        }
+        return true
       }
+      await pollFor(allChecksOk)
+      expect(allChecksOk()).toBe(true)
```
pzduniak: Since we're in JS rather than iced, why not throw a timeout error in pollFor? I think that'd be more intuitive, I nearly pushed broken tests to master in iced because I didn't realize that pollFor will "fail silently". If you keep it as-is for consistency across codebases then I'll totally understand it, it's just a suggestion.

malgorithms: oh, that's fine. This has nothing to do with iced. I just figured the test would fail either way if the poll timed out. But I'll change it.

pzduniak: Idea #2 - make `pollFor` return the final value, so this would become

`expect(await pollFor(allChecksOk)).toBe(true)`

Still won't require duplicate checks after pollFor while making the code clearer as the syntax itself will describe what it's doing vs just the name.

malgorithms: I just went with option 1 before I saw idea 2. Option 1 seems fine - this is just a little helper fn for tests anyway.

malgorithms:
@pzduniak if you wanna give me the thumb  I'll merge

PR #137, opened 2019-04-12
team inviter demo

pzduniak:
@patrickxb all using existing APIs!

### DIFF
--- a/demos/es7/team-inviter.js
+++ b/demos/es7/team-inviter.js
@@ -1,0 +1,63 @@
+#!/usr/bin/env node
+const Bot = require('../../index.js')
+
+const bot = new Bot()
+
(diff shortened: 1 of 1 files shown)


pzduniak:
Live demo running as https://keybase.io/kielbase - just dm the bot!

pzduniak:
@malgorithms I'm thinking maybe we could merge this in to showcase the new team API?

malgorithms:
@pzduniak looks great to me!  go for it.

A bonus to the script would be if the bot could add them to multiple channels at once. I think we need to add API support for that though.

pzduniak:
Do you think I should file a ticket for this? It doesn't seem that complicated.

PR #139, opened 2019-05-09
auto log send on crash

pzduniak:
requires https://github.com/keybase/client/pull/17460 which currently serves as a demo as it crashes the service after 5s

I didn't generate docs etc. but it's more of a proposal / starting point for a discussion rather than a full solution. I don't like how I had to do some of this stuff but there's really no other way to attribute log sends to usernames. I'm also curious why service runs with `-d` if serviceLogFile is passed. Is anyone using this?


pzduniak:
I might as well add it as a part of this ticket since it already has a keybase/client component.

pzduniak:
Added feedback msg support, wdyt?

malgorithms:
LGTM

pzduniak:
merge pending client review

PR #144, opened 2019-06-11
Add showLocal support, demo with a flip triggered by the bot

pzduniak:


pzduniak:
I think this one might be test-worthy? wdyt?

PR #146, opened 2019-06-13
converting to typescript

malgorithms:
ping @chrisnojima


package.json
```diff
@@ -34,59 +32,53 @@
...
   },
   "lint-staged": {
-    "*.js": [
-      "eslint --fix",
-      "flow focus-check"
+    "*.ts": [
+      "eslint --fix"
     ],
-    "*.{js,json,md}": [
+    "*.{json,md}": [
```
nathunsmitty: I think we want to run prettier on our ts files too

.eslintignore
```diff
@@ -1,3 +1,4 @@
 /index.js
 /index.js.map
 flow-typed
+lib
```
nathunsmitty: 👍 

May want to make this `/lib` just because `lib` is a pretty common name. We could also remove `flow-typed`.

__tests__/chat.test.ts
```diff
@@ -1,28 +1,37 @@
...
-import config from './tests.config.js'
+import config from './tests.config'
 import {timeout} from '../lib/utils'
 import {pollFor} from './test-utils'
 import {promisify} from 'util'
+import {ChatChannel, ReadResult, MessageSummary} from '../lib/chat-client/types'
+import {OnMessage} from '../lib/chat-client'
 
-test('Chat methods with an uninitialized bot', () => {
+test('Chat methods with an uninitialized bot', (): void => {
```
nathunsmitty: Does TS give an error if we don't include the `: void` type signature on functions like these? Would of thought that it could infer that 🤷‍♂

__tests__/chat.test.ts
```diff
@@ -202,7 +220,7 @@ describe('Chat Methods', () => {
         },
         {name: `${config.bots.alice1.username},${config.bots.bob1.username}`},
       ]
-      const okChecks = []
+      const okChecks: any[] = []
```
nathunsmitty: can `okChecks` be typed as a boolean array?

lib/chat-client/index.d.ts
```diff
@@ -0,0 +1,213 @@
+import ClientBase from '../client-base';
```
nathunsmitty: wait, aren't we supposed to ignore `lib/`?

src/chat-client/index.ts
```diff
@@ -0,0 +1,465 @@
+// @flow
```
nathunsmitty: Don't think we need this line!

demos/math-bot.js
```diff
@@ -11,6 +11,7 @@ const mathjs = require('mathjs')
 //          /math sqrt(pi/2) * 3!`
 //
 
+console.log(Bot)
```
nathunsmitty: do we want to keep this addition?

jest.config.js.old
```diff
@@ -0,0 +1,179 @@
+// For a detailed explanation regarding each configuration property, visit:
```
nathunsmitty: Do we want to keep this file around?

__tests__/chat.test.ts
malgorithms: something was mad and made me add it...maybe eslint?

package.json
```diff
@@ -34,59 +32,53 @@
...
       "prettier --write",
       "git add"
     ]
   },
   "scripts": {
     "docs": "documentation readme lib/** --config documentation.yml --section=API",
-    "build": "rollup -c",
-    "dev": "rollup -c --watch",
+    "tsc": "tsc",
```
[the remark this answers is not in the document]
malgorithms: ok I'll go back to those

package.json
malgorithms: yup, will add .ts

.eslintignore
malgorithms: oh nice catch

__tests__/chat.test.ts
malgorithms: yeah I think so

PR #148, opened 2019-06-17
docs: Add a readme section about test accounts

nathunsmitty:
Adds a section about how tests work and how to run them in the readme. Also fixes an eslint complaint and makes clear that the `alice` account needs two paperkeys.


README.md
```diff
@@ -44,9 +44,7 @@ bot
 
     const channel = {name: 'kbot,' + bot.myInfo().username, public: false, topicType: 'chat'}
     const message = {
-      body: `Hello kbot! This is ${bot.myInfo().username} saying hello from my device ${
```
nathunsmitty: These changes are from prettier, due to the fact that we increased the line width to 140 characters.

PR #156, opened 2019-07-08
Coyne/logmode

malgorithms:
a new option that has the bot copy its service files to a directory of your choosing, and also to add more verbose logging alongside that.


src/index.ts
```diff
@@ -56,9 +69,21 @@ class Bot {
...
+    await this._maybePrepDebugDir(options)
     await this._service.initFromRunningService(homeDir, options)
     await this._initSubBots(options)
+    await this._maybeLogCopyLoop(options)
+    this.debugLog('initialized')
+  }
+
+  private _beginInitState(): void {
+    if (this._initStatus !== 'preinit') {
+      throw new Error(`tried to init, but state is already ${this._initStatus}`)
```
nathunsmitty: Do we want to allow a deinited bot to be reinitialized? I think I remember being able to do that, but this check no longer allows that.

src/index.ts
```diff
@@ -81,10 +106,56 @@ class Bot {
...
+   * This function is also called automatically when in adminDebug mode, and will record plaintexts of messages into that directory,
+   * so use wisely.
+   * @memberof Bot
+   * @example
+   * bot.log("My bot is ready to go.")
+   */
+  public async debugLog(text: string, infoOrErr?: 'I' | 'E'): Promise<void> {
+    if (this.adminDebugDirectory) {
+      const code = infoOrErr ? infoOrErr : 'I'
+      const line = `${new Date().toISOString()} [${code}] ${text}\n`
```
nathunsmitty: We should use `os.EOL` here for windows compatibility. https://nodejs.org/api/os.html#os_os_eol

package.json
```diff
@@ -41,11 +41,15 @@
...
   "scripts": {
+    "modules": "yarn --frozen-lockfile --prefer-offline",
+    "modules-prod": "yarn --frozen-lockfile --prefer-offline --prod",
     "docs": "documentation readme lib/** --config documentation.yml --section=API",
-    "build": "tsc",
     "dev": "tsc --watch",
     "test": "jest --runInBand",
-    "release": "standard-version"
+    "release": "standard-version",
+    "build": "rm -rf lib/ && tsc",
```
nathunsmitty: I hear often that `rm` won't work on windows; a common suggestion is the use [rimraf](https://www.npmjs.com/package/rimraf), which is cross-platform. Not sure how relevant this advice still is given things like WSL but figured I'd mentioned it.

src/index.ts
```diff
@@ -81,10 +106,56 @@ class Bot {
...
+    }
+  }
+
+  /**
+   * If bot is initialized with an optional directory `adminDebugDirectory`, this will write that string into a file in there
+   * This function is also called automatically when in adminDebug mode, and will record plaintexts of messages into that directory,
+   * so use wisely.
+   * @memberof Bot
+   * @example
+   * bot.log("My bot is ready to go.")
```
nathunsmitty:
```suggestion
   * bot.debugLog("My bot is ready to go.")
```

src/index.ts
```diff
@@ -25,11 +31,13 @@ class Bot {
...
   public constructor() {
-    this._workingDir = randomTempDir()
-    this._service = new Service(this._workingDir)
-    this.chat = new ChatClient(this._workingDir)
-    this.wallet = new WalletClient(this._workingDir)
-    this.team = new TeamClient(this._workingDir)
+    this._botId = crypto.randomBytes(16).toString('hex')
+    this._workingDir = path.join(os.tmpdir(), `keybase_bot_${this._botId}`)
+    this._service = new Service(this, this._workingDir)
+    this.chat = new ChatClient(this, this._workingDir)
```
nathunsmitty: I'm not sure I'm a big fan of passing a reference of the entire bot to each of the subclients – it creates circular references and starts to break down the distinction between the roles of the clients and the core bot. Could we instead just pass the `debugLog` file path or function to the clients?

src/index.ts
malgorithms: no, I removed that on purpose and changed the test case. I think it's dangerous, and no one really needs to do it in practice. While I don't think we have any bugs in our library, it seems a bit weird to me to have the bot possible to cross from one user to another. Safer to instantiate a new one.

src/index.ts
malgorithms: on it

package.json
malgorithms: I think I'd rather not support windows than import another library right now. we can make a separate ticket out of windows support later. I think there are a bunch of reasons this won't work on windows, not just the building.

src/index.ts
malgorithms: funny, I started that way, and IMO it gets uglier to go the other way, as things start piling on (bot's working directory, bot's id, bot's log function, etc.)...what's wrong with a circular reference? this seems like the kind of place one makes sense.

malgorithms: I understand a lot of the issues with how people use circular references in shitty ways, but these sub-bot clients are 1-1 with their Bot class, and in fact are basically part of the same instance.

hmm, will think about this more.

malgorithms: you know...ok...gonna do it your way for now. I may come back to this if more things get tacked on.

nathunsmitty: Ok. I think it'd definitely be worth reevaluating the bot architecture sometime, especially since it's grown a lot since it was first designed.

malgorithms: ok, changed it up, lmk. I'd like to merge this soon.

PR #157, opened 2019-07-09
Mark lib/ directory as linguist-generated

nathunsmitty:
This should make the GitHub diffs easier to read by hiding the generated files of `lib/` by default. We do the same thing in the client repo for the generated RPC types!


malgorithms:
oooh cool! merge that!

PR #160, opened 2019-07-15
node-which instead of sh -c which, allow passing the binary path directly

pzduniak:
working on a serverless template...


pzduniak:
getting all this stuff out will be a long process but that's all of the changes required in keybase-bot

i re-ran `yarn`, looks like there was lots of updated deps, mostly dev ones

node-which is included in npm and only depends on isexe which is written by the same author - I went through the code and it looks perfectly fine. Should be quite secure if npm is using it!

PR #161, opened 2019-07-17
Update dependencies

nathunsmitty:
Also moves `@typescript-eslint/parser` to dev dependencies


malgorithms:
sweet thanks for doing this. consider this thumbed when you're ready...also maybe bump package.json

nathunsmitty:
@malgorithms I've got some other changes coming, so I think it'd make the most sense to bump package.json when those are done

PR #163, opened 2019-07-18
Add types field for use by typescript apps

nathunsmitty:
Adds a `types` entry to the package.json so that typescript apps that use this package can take advantage of our types. See https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package for details.

I was really confused about why my bot import was `any` and this fixes that haha

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
     "gpg"
   ],
   "main": "lib/index.js",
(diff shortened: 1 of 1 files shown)


malgorithms:
awesome go for it

PR #165, opened 2019-07-25
Commands advertisement support

pzduniak:
not sure what's our approach for docs?


demos/es7/advertised-echo.js
```diff
@@ -0,0 +1,60 @@
...
+        }
+      ],
+    })
+
+    const onMessage = async message => {
+      if (message.content.type !== 'text') {
+        return
+      }
+
+      if (!message.content.text.body.startsWith('!echo ')) {
```
nathunsmitty: Would it make sense to modify the watch methods so that they automatically ignore non-command messages and ignore the prefixes?

__tests__/chat.test.ts
```diff
@@ -508,6 +508,36 @@ describe('Chat Methods', (): void => {
...
+        channel: channel,
+      })
+      expect(list.commands).toContain({
+        name: '!helloworld',
+        description: 'sample description',
+        usage: 'test',
+        username: alice1.myInfo().username,
+      })
+    })
+  })
```
nathunsmitty: Idea: clear commands and check that `list.commands` is empty here

demos/es7/advertised-echo.js
malgorithms: hmm, what if the bot registers commands but also has access to all messages? I'd say either it should work like it does here, or, we add a pretty wrapper around doing this extra work and doesn't complicate the inside of any of our existing `watch` functions.

so here you could have:

```
await bot.chat.clearCommands()
await bot.chat.advertiseCommands(/*stuff*/)
bot.chat.watchForCommand('!echo', (message, argumentText) => {
 bot.chat.Send(message.channel, body: argumentText)
});
```

watchForCommand does the filtering you want, both to make sure it's only getting the command you ask for, and also it provides a convenient `argumentText`.

lib/chat-client/index.d.ts
```diff
@@ -157,6 +157,27 @@ declare class Chat extends ClientBase {
...
+     * @param advertisement - advertisement parameters
+     * @example
+     * // check demos/es7/advertised-echo.js
+     */
+    clearCommands(): Promise<void>;
+    /**
+     * Lists all published command advertisements, either all or for a certain channel.
+     * @param lookup - either conversation id or channel
+     * @example
+     * // check demos/es7/advertised-echo.js
```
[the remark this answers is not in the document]
pzduniak: fair enough i admit i just halfassed this one

demos/es7/advertised-echo.js
pzduniak: counterpoint: we don't want the base library to contain a command parser

malgorithms: my proposal above doesn't do anything to the messages it gets from watchForCommand, so someone could do whatever they want and ignore argumentText. The convenience is just saving 95% of command authors from rewriting the same garbage every time where the (1) check to make sure it's a text message, (2) check to make sure it starts with their command, (3) perform a substring extraction on their command. It would be optional to use, just a pretty function. 

in any case, my proposal doesn't change anything about what you have here, since it's just something new we can add.

pzduniak: weirdly enough i can't see your previous comment here, i still think we should do it in another module and make it reusable

PR #169, opened 2019-08-21
detached service support

malgorithms:
@patrickxb this change makes it so the keybase bot library can spawn its keybase service as `detached` which means SIGINT won't get propagated to it. normally we don't want this, because if a user just hits ctrl-c to kill a bot they write, and don't call deinit(), then their service can get stuck running. So we want all the spawned processes to follow their parent.

but this allows me to do what you wanted, which is have a sigint hit the parent, catch that, then do some final stuff, then call deinit() which will shut down everything including the detached processes.

let me know if it looks good and I'll add it to the bot.


roychowdhuryrohit-dev:
This option only works with ctrl-c but not with PM2 when it sends SIGINT.
```
Shutting down server.
Error: Login required
    at new ErrorWithCode (/workspaces/social-graph/node_modules/keybase-bot/lib/client-base/index.js:63:28)
    at Chat.<anonymous> (/workspaces/social-graph/node_modules/keybase-bot/lib/client-base/index.js:141:35)
    at step (/workspaces/social-graph/node_modules/keybase-bot/lib/client-base/index.js:46:23)
    at Object.next (/workspaces/social-graph/node_modules/keybase-bot/lib/client-base/index.js:27:53)
    at fulfilled (/workspaces/social-graph/node_modules/keybase-bot/lib/client-base/index.js:18:58)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) {
  code: 0
}
```

src/utils/options.ts
```diff
@@ -14,4 +14,7 @@ export interface InitOptions {
   adminDebugDirectory?: string
   // Overrides the path to the keybase binary
   keybaseBinaryLocation?: string
+  // With this turned on, a SIGINT won't kill your Keybase service,
+  // but you must remember to call deinit() or it will survive
+  useDetachedService?: boolean
```
roychowdhuryrohit-dev, line 19: Doesn't work when PM2 sends SIGINT.

PR #177, opened 2019-09-09
fix ts docs - trying out 2nd approach

pzduniak:


pzduniak:
This generates mostly OK docs. Requires https://github.com/documentationjs/documentation/pull/1278

Still don't know if `shallow` is required or not, looks like its failing to generate docs for `chat1.` etc.

nathunsmitty:
Okay, I think this is ready for a review. Docs are generated for all bot functions.

The most important asterisk is that that documentation is not generated for any of the generated types. This is for two reasons:
1. Many of the generated types don't have documentation specified in the AVDL files, and thus no documentation is included in the generated files.
2. There are a handful of internal/server types (e.g., `PGPSigVerification`) that aren't used by the bot that do have documentation, which adds noise to the README. I couldn't find a way to exclude types on a per type basis, only on a per file basis.

Number 1 is a pretty easy fix, I'm happy to add some documentation lines to existing avdl definitions. Number 2 is a little harder; the best solution would be to exclude these types from the generated bot types completely. The best way to do that would likely be to define a new namespace for the api types (`api1` or similar) in the avdl files and have the avdl compiler compile only those types and any types they depend on.

pzduniak:
I can't review because it's my own PR but I think this is better than what we have right now with all the outdated docs, so feel free to merge.

PR #180, opened 2019-09-13
Coyne/let back in WIP

malgorithms:
This PR is just to add 2 small features to the chat bot, 

1. a call to get any reset convos that the bot has to let back in:

```
const problemChats = await bot.chat.getResetConvMembers()
```

and

2. actual calls to let them back in:

```
await bot.chat.addResetConvMember(p)
```

Coming along for the ride are newly compiled types (due to changes in keybase/client side to support this and other things they're working on). So the PR looks bigger at first glance.

Also added a couple TypeScript demos - one demonstrating these new calls.

cc @mmaxim


__tests__/team.test.ts
```diff
@@ -1,28 +1,28 @@
...
+import {TeamRole, TeamDetails} from '../lib/types/keybase1'
 
 test('Team methods with an uninitialized bot', (): void => {
   const alice1 = new Bot()
   const options = {team: config.teams.alicesPlayground.teamname}
   expect(alice1.team.listTeamMemberships(options)).rejects.toThrowError()
 })
 
-function pluralizeRole(r: TeamRole): TeamRolePlural {
+function pluralizeRole(r: TeamRole): 'owners' | 'admins' | 'readers' | 'writers' {
```
chrisnojima: don't have to change this but as an FYI you should be able to 
1. Mostly infer most return types
1. Use `as const` to stop TS from type widening. Aka it'll see `return 'owners' as const` as the literal `'owners'` and not widen the type to `string`

src/chat-client/index.ts
```diff
@@ -480,6 +480,38 @@ class Chat extends ClientBase {
...
+  /**
+   * Let's a conversation partner back in after they've reset their account. You can
+   * get a list of such candidates with getResetConvMembers()
+   * @param username - the username of the user who has reset
+   * @param conversationId
+   * @example
+   * await bot.chat.addResetConvMember({username: "chris", conversationId: "abc1234567"})
+   */
+  public async addResetConvMember(param: chat1.ResetConvMemberAPI): Promise<chat1.GetResetConvMembersRes> {
+    await this._guardInitialized()
```
chrisnojima: unrelated to this PR but
1. TS has `protected` / `private' etc so using underscore prefix isn't necessary
1. This function can throw and its not wrapped in a try so that'll bubble up (is that intended?)

malgorithms: 1. cool - `_guardInitialized` was already `protected`; I'm throwing JS (non-Typescript) people a bone by making our protected and private members underscore named to hint.

2. yup, intended. Most of our API calls throw their errors up if there's a problem; scripts are expected to catch them if there's an issue.

__tests__/team.test.ts
malgorithms: As long as I'm sticking with explicit return types everywhere (currently using that setting), the `as const` thing is unnecessary, right? I don't think this widens to `string` as it is.

PR #194, opened 2019-11-22
JS bindings for `keybase api kvstore`

cjb:
r? @mmou @xgess CC @joshblum 

This PR's a bit hard to scan through because in this repo we commit the compiled TS->JS as well, rather than just the source TS.  There are three new source files:

* `src/kvstore-client/index.ts`: The JS bindings for kvstore.

* `__tests__/kvstore.test.ts`: A regression test for kvstore, brought over from the Golang binding tests.

* `demos/typescript/team-kvstore-1.ts`: An example bot that responds to kvstore commands, brought over from a sample bot in an earlier commit of @mmou's Golang PR.  It does a little more than the Golang version did, by printing a description of the outcome rather than just the verbatim API response.

I know the latter two files seem redundant: they're very similar code.  But I think they're both worth having -- the test functions as a regression test, and the demo bot functions as a template for bot authors to start from and get quickly up to speed with deploying kvstore bots.  Kudos to @mmou for some great sample code to work from!  This was fun to work on.

I can't seem to get the documentation to regenerate (even just against master) without seeing extra indentation be added and the first `- Bot` section get moved to the bottom.  @pzduniak, @nathunsmitty, etc, any ideas?  Does `yarn docs` result in a large diff for you too, on master?  How about on this branch?  Perhaps the version of `documentation` currently in `package.json` isn't the one that generated the current docs?


lib/kvstore-client/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+    isDeleted(res: keybase1.KVGetResult): boolean;
+    /**
+     * Determine whether the result of a `get()` call describes an entryKey that has an existing value.
+     * @memberof KVStore
+     * @param res - The `get()` result to determine the status of.
+     * @returns - Whether this key's value is present.
+     * @example
+     * bot.kvstore.isPresent(res).then(isPresent => console.log({isPresent}))
+     */
+    isPresent(res: keybase1.KVGetResult): boolean;
```
xgess: 👍

mmou:
given that none of the other demo examples are numbered, and that the number 1 doesn't mean much outside of this repo, perhaps we should rename the file `team-kvstore-1.ts` to `team-kvstore.ts`

demos/typescript/team-kvstore-1.ts
```diff
@@ -0,0 +1,189 @@
...
+      const res = await this.bot.kvstore.listNamespaces(this.teamname)
+      const message = res.namespaces.join(', ')
+      await this.sendMessage(message || 'There are no namespaces.')
+    } catch (err) {
+      await this.sendMessage(`${err}`)
+    }
+  }
+
+  async handleListEntryKeys(body: string): Promise<void> {
+    // !storage list someteam somenamespace
```
mmou: nit: inconsistent documentation for expected command string format

__tests__/kvstore.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    const res = await alice1.kvstore.get(team, namespace, entryKey)
+    expect(res).toEqual(getResultMatcher)
+    expect(res.entryValue).toEqual('value1')
+    expect(res.revision).toEqual(rev)
+  })
+
+  it('Cannot delete with a stale revision', (): void => {
+    // Increment rev so that later tests always have a usable rev value prepared.
+    rev += 1
+
```
mmou: delete extra line

__tests__/kvstore.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+
+  it('Gets values', async (): Promise<void> => {
+    const res = await alice1.kvstore.get(team, namespace, entryKey)
+    expect(res).toEqual(getResultMatcher)
+    expect(res.entryValue).toEqual('value1')
+    expect(res.revision).toEqual(rev)
+  })
+
+  it('Cannot delete with a stale revision', (): void => {
```
mmou: looks like this test is actually testing that you can't delete with a future, not stale, revision

cjb:
@xgess @mmou Ready for another look, thanks!

PR #198, opened 2020-01-05
add watchForNewConversation

songgao:
This allows the bot to say hi when added into a conversation. I refactored the `_chatListen` method so it can be reused for the other listen.

cc @keybase/hotpotatosquad


src/chat-client/index.ts
```diff
@@ -661,29 +658,102 @@ class Chat extends ClientBase {
...
         }
       }
     }
-    lineReaderStdout.on('line', onLine)
+    this._adminDebugLogger.info(`spawningChatListenChild on channel=${JSON.stringify(channel || 'ALL')}`)
+    this._spawnChatListenChild(args, onLine)
+  }
+
+  /**
+   * Spawns the chat listen process for new channels and handles the calling of onConv, and onError.
```
joshblum: This listener will also receive new message notifications, why not just have a single listener that can take `onMessage` and `onConv` callbacks?

songgao: The JS lib relies on the Go side to filter on channels. So we already have multiple api-listen calls happening if you watch multiple channels, or watch all messages and a particular channel at the same time. I guess we could use the same one for watching all messages and new conversations, and keep the channel watching the same way as before, but it didn’t seem worth it to me if it’s not the same form for everything.

joshblum: Gotcha makes sense!

PR #200, opened 2020-01-10
Bump to version 3.2.0

joshblum:


joshblum:
@malgorithms Can you publish when you get a sec?

PR #203, opened 2020-01-15
fix notification-settings

songgao:
This lib stopped working seemingly after https://github.com/keybase/client/pull/21986 . So just use the `=<Boolean>` syntax for that.


joshblum:
@songgao this should have been fixed in https://github.com/keybase/keybase-bot/pull/201, are you sure you were running the right version? Not sure if 3.2.0 is published yet

songgao:
> @songgao this should have been fixed in #201, are you sure you were running the right version? 

I think so -- I was running against master and it didn't work. Look at the diff; it's based on a commit that has the change from that PR.

joshblum:
I see in the diff, just wasn't sure what your test setup was

PR #219, opened 2020-02-19
feat(chat client): Adding exploding lifetime option to send and attach.

jhoscar1:


src/chat-client/index.ts
```diff
@@ -48,6 +48,7 @@ export interface ChatSendOptions {
   membersType?: chat1.ConversationMembersType
   confirmLumenSend?: boolean
   replyTo?: chat1.MessageID
+  explodingLifetime?: string
```
songgao: Thanks for the PR! I think `explodingLifetime` should be a number here (same for the other one), as in millisecond, just like how the duration is specified in other JS apis like `setTimeout`. Then before sending it to the CLI api, we just need to add the `ms` so they can be properly [parsed](https://golang.org/pkg/time/#ParseDuration) on the [Go side](https://github.com/keybase/client/blob/b9875b85992d7af0573cab8afb6aa46f7227997a/go/client/chat_api_handler.go#L167).

jhoscar1:
@songgao thanks for the quick response! Will make these changes

jhoscar1:
I wasn't certain if the options processing should happen in the formatAPIObjectInput function but it seemed like a good single point to place it @songgao

pzduniak:
https://github.com/keybase/keybase-bot/blob/master/src/kvstore-client/index.ts#L75

I'd implement it like this instead, possibly do something like `explodingLifetime: options?.explodingLifetime ? `${options.explodingLifetime}ms` : undefined`. That API formatting code is ugly enough as-is.

jhoscar1:
@pzduniak Good call. Thanks! Moved the suffix addition to the client methods.

src/utils/formatAPIObject.ts
```diff
@@ -107,10 +103,7 @@ function matchBlacklist(context: FormatAPIObjectOutputContext | null): boolean {
  * @param key - The key to apprent to the parent array.
  * @returns - A new context.
  */
-function buildContext(
-  context: FormatAPIObjectOutputContext | null,
-  key: any
-): FormatAPIObjectOutputContext | null {
+function buildContext(context: FormatAPIObjectOutputContext | null, key: any): FormatAPIObjectOutputContext | null {
```
songgao: Mind revert whitespace changes in this file?

src/chat-client/index.ts
```diff
@@ -309,7 +312,12 @@ class Chat extends ClientBase {
   ): Promise<chat1.SendRes> {
     await this._guardInitialized()
     const conv = this.getChannelOrConversationId(channelOrConversationId)
-    const args = {...options, ...conv, filename}
+    const args = {
+      ...options,
+      explodingLifetime: options.explodingLifetime ? `${options.explodingLifetime}ms` : null,
```
songgao: s/null/undefined. Same above.

src/utils/formatAPIObject.ts
jhoscar1: @songgao I'm not sure what I'm doing wrong but eslint keeps overwriting these files

songgao:
Thanks @jhoscar1 !

PR #231, opened 2020-03-28
Correct bot advertisements to demo restricted bots

pzduniak:
Looks like just the advertisement syntax was broken. Otherwise the demo works fine.


pzduniak:
Not bothering with a review, merging it outright since it's just a typo - the PR was in place just in case I were to find more bugs.

PR #232, opened 2020-04-01
improve docker example

stasbar:
Hi,
Currently, the example of Dockerfile is error-prone.
When you try to run it with typescript, which generates `/dist` it will fail due to Permission access error.
After some investigation, I've figured out that `keybaseio/client:nightly-node-slim` leave the image with `keybase` user. Command `WORKDIR /app` creates the `/app` directory at root level access, which prevents from creating there `/dist` directory.
My proposal is to set `WORKDIR` to `/home/keybase` (where the user has write access)  by default.
Additionally 
```
COPY package*.json ./
RUN npm install # or use yarn
COPY . .
```
for free performance improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,8 @@ main()
 
 ```dockerfile
 FROM keybaseio/client:nightly-node
-WORKDIR /home/keybase
+RUN mkdir /app && chown keybase:keybase /app
+WORKDIR /app
 COPY package*.json ./
 RUN npm install # or use yarn
 COPY . .
--- a/README.md
+++ b/README.md
@@ -170,9 +170,10 @@ main()
 
 ```dockerfile
 FROM keybaseio/client:nightly-node
-WORKDIR /app
-COPY . /app
+WORKDIR /home/keybase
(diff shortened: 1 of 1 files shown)


pzduniak:
>Command WORKDIR /app creates the /app directory at root level access, which prevents from creating there /dist directory.
Great find, that's a bug in the Keybase image. I'll fix it first, then I'll revisit your PR. Thanks!

README.md
```diff
@@ -170,9 +170,10 @@ main()
 
 ```dockerfile
 FROM keybaseio/client:nightly-node
-WORKDIR /app
-COPY . /app
+WORKDIR /home/keybase
```
pzduniak, line 173:
```suggestion
RUN mkdir /app && chown keybase:keybase /app
WORKDIR /app
```

stasbar: Yeah, you are right with "Keybase would dump all of its cache and config files in that volume", your suggestion is better.

pzduniak:
Thank you for your contribution!

PR #237, opened 2020-04-22
kvstore change empty to nil

xgess:
* update the types to the most recent keybase client release
* support new change to kvstore entryValue behavior (`null` instead of `''` when unset)

suggest reviewing this by commit
and perhaps not publishing until the keybase client 5.5.0 goes out

depends on https://github.com/keybase/client/pull/23884


xgess:
@keybase/picnicsquad

src/kvstore-client/index.ts
```diff
@@ -143,7 +146,7 @@ class KVStore extends ClientBase {
    * bot.kvstore.isDeleted(res).then(isDeleted => console.log({isDeleted}))
    */
   public isDeleted(res: keybase1.KVGetResult): boolean {
-    return res.revision > 0 && res.entryValue === ''
+    return res.revision > 0 && !entryValueIsSet(res.entryValue)
```
mmou, line 149: just to understand, do we expect entryValue==undefined if the entry is deleted?

src/kvstore-client/index.ts
```diff
@@ -12,6 +12,9 @@ export const ErrorIsWrongRevision = (error: ErrorWithCode): boolean => error.cod
 export const ErrorIsBadGeneration = (error: ErrorWithCode): boolean => error.code === KVStoreErrorType.BadGeneration
 export const ErrorIsNotFound = (error: ErrorWithCode): boolean => error.code === KVStoreErrorType.NotFound
 
+const entryValueIsSet = (entryValue: string | null | undefined): boolean =>
+  entryValue !== '' && entryValue !== null && entryValue !== undefined
```
mmou, line 16: should we keep `entryValue !== ''` or not?

src/kvstore-client/index.ts
xgess: `undefined` is what would happen if the key were missing from the deserialized json coming out of the client (as opposed to explicitly being set to null, which is what it should be after the client PR is merged). it would be unexpected for the client to leave it out, but i'd say not irrational for the bot library to insulate itself from the client potentially doing this.

src/kvstore-client/index.ts
xgess: i added this to the python library. yes, i think we should leave it to make the bot libraries a little more robust to client versioning.

